PR #3, opened 2023-09-20
change package urls, added vite-plugin dependency

jfboeve:


package.json
```diff
@@ -58,6 +58,7 @@
   },
   "dependencies": {
     "@lightningjs/threadx": "^0.3.0",
+    "@lightningjs/vite-plugin-import-chunk-url": "^0.3.0",
```
frank-weindel, line 61: This is only a dependency for apps that use the renderer. The renderer itself does not use this plugin (nor vite for that matter).

frank-weindel (approved):

PR #7, opened 2023-09-26
fix: Update code to make it work on Samsung Tizen 3.5

el-gringo:
- Removes the use of FinalizationRegistry
- Use a real canvas instead of OffscreenCanvas

Note: Need disable `build.minify` in `vite.config.js`

### DIFF
--- a/src/core/text-rendering/renderers/CanvasTextRenderer.ts
+++ b/src/core/text-rendering/renderers/CanvasTextRenderer.ts
@@ -84,12 +84,18 @@ export interface CanvasTextRendererState extends TextRendererState {
 }
 
 export class CanvasTextRenderer extends TextRenderer<CanvasTextRendererState> {
+  /*
   protected canvas: OffscreenCanvas;
   protected context: OffscreenCanvasRenderingContext2D;
+*/
+  protected canvas: HTMLCanvasElement;
+  protected context: CanvasRenderingContext2D;
 
   constructor(stage: Stage) {
     super(stage);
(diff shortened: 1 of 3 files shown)


chiefcll:
Thanks - we'll need to figure out flags for these pieces to work on older devices

frank-weindel:
Thanks for the PR!

- OffscreenCanvas in CanvasTextRenderer: We could possibly add something that detects if Core is running on the main thread and then always use the HTML Canvas when thats the case.
- FinalizationRegistry: This is needed for the texture memory system. We may need to adopt a model that simply counts what textures are actively assigned to nodes that are attached to render tree.

chiefcll:
@frank-weindel - just tested on xi6 boxes and we also dont have FinalizationRegistry - is there a polyfill or can we remove this for now?

frank-weindel:
Those devices will eventually receive the WPE versions we were targeting during development. There's no polyfill for FinalizationRegistry so another layer of memory management that doesn't rely on it will need to be developed.

jonybekov:
WebOS versions up to 5 all also will not work because of FinalizationRegistry (Chrome 84+) and OffscreenCavas (Chrome 69+).
image
 
 Support for older browsers would be awesome! (at least with polyfills).

chiefcll:
Solid will need to target Chrome 49+

wouterlucas:
Closing this in favor of https://github.com/lightning-js/renderer/issues/26

PR #16, opened 2023-09-28
Clipping

frank-weindel:
Clipping is enabled by setting the `clipping` property to true on a Node.

Clipping while a node is rotated or scaled currently leads to undesired results. More work with transforms in general is needed before we can improve how it works with clipping.

The `clipping` example tests various clipping scenarios. A new test example utility component PageContainer along with a special function called paginateTestRows() makes it easy to create multi-row / multi-page test examples which we can use in the future to create regression snapshots.

Upgrade to threadx 0.3.1 to fix issue where text does not work in ThreadX mode.

Screen Shot 2023-09-28 at 11 33 21 AM


frank-weindel:
I'm aware of a misalignment issue between rendered quad bounds and their clip bounds that sometimes appears. I've spent a lot of time on it already and am continuing to look into it. It's a bit of a head scratcher and I think is related to rounding differences between CPU and GPU.

chiefcll:
Sounds about right (pretending to understand that comment) - I'm checking out clipping now in Solid

chiefcll (approved):
Working!

PR #18, opened 2023-09-29
alpha to RenderOp with automatic shader props

jfboeve:
I've added alpha to the shader props which can be used for automatic shader props similar to [$dimensions](https://github.com/lightning-js/renderer/pull/1). This is a prop we need  for some of the DynamicShader effects.

This also pull request is related to #15


src/core/renderers/webgl/shaders/DynamicShader.ts
```diff
@@ -314,7 +314,7 @@ export class DynamicShader extends WebGlCoreShader {
 
   static override resolveDefaults(
     props: DynamicShaderProps,
-  ): Required<DynamicShaderProps> {
+  ): Record<string, unknown> {
```
frank-weindel, line 317: Is this necessary?

src/core/renderers/webgl/WebGlCoreShader.ts
```diff
@@ -42,7 +42,7 @@ import { isWebGl2 } from './internal/WebGlUtils.js';
  * Shader's who's rendering depends on the dimensions of the Node being rendered
  * should extend this interface from their Prop interface type.
  */
-export interface DimensionsShaderProp {
+export interface AutomaticShaderProps {
```
frank-weindel, line 45: There should be a new interface for the automatic alpha prop. You can have `DynamicShaderProps` implement multiple interfaces in order to pass both props (and all the documentation tied to them) up. This allows each shader to define specifically which automatic props it requires.

src/core/renderers/webgl/shaders/RoundedRectangle.ts
```diff
@@ -65,7 +65,7 @@ export class RoundedRectangle extends WebGlCoreShader {
 
   static override resolveDefaults(
     props: RoundedRectangleProps,
-  ): Required<RoundedRectangleProps> {
+  ): Record<string, unknown> {
```
frank-weindel, line 68: Same as above. Is there a reason for this? The point of having the return type be specific is so TypeScript can tell us if we don't fill out all the proper defaults.

src/core/renderers/webgl/shaders/RoundedRectangle.ts
```diff
@@ -81,7 +81,7 @@ export class RoundedRectangle extends WebGlCoreShader {
     gl.bindTexture(gl.TEXTURE_2D, textures[0]!.ctxTexture);
   }
 
-  protected override bindProps(props: Required<RoundedRectangleProps>): void {
+  protected override bindProps(props: RoundedRectangleProps): void {
```
frank-weindel, line 84: bindProps should always be getting the props after defaults were applied.

frank-weindel (changes requested):
See comments

src/core/renderers/webgl/WebGlCoreShader.ts
jfboeve: Alright I've reverted that interface back to DimensionsShaderProp, and added AlphaShaderProp. This resolves the Record<string, unknown> and Required Issues mentioned in other comments too.

frank-weindel (approved):

src/core/renderers/webgl/WebGlCoreRenderer.ts
```diff
@@ -491,10 +494,12 @@ export class WebGlCoreRenderer extends CoreRenderer {
       if (recursive) {
         throw new Error('Unable to add texture to render op');
       }
+
       const { shader, shaderProps, dimensions, clippingRect } = curRenderOp;
       this.newRenderOp(
         shader,
         shaderProps,
+        alpha
```
frank-weindel, line 502: @jfboeve missing a comma. This is causing a TS error

frank-weindel (changes requested):
See comment. There's a TS error

src/core/renderers/webgl/WebGlCoreRenderer.ts
chiefcll: dangling commas - should the linter auto fix this?

frank-weindel: Nah this was a legit syntax error

jfboeve: Sorry must have gone wrong while merging, also alpha needed to be extracted from the curRenderOp.

PR #27, opened 2023-10-03
Replace FinalizationRegistry based Texture Usage GC with Manual Strategy

frank-weindel:
- Usaged-Based Texture Memory management now defaults to a new manual reference count and time threshold strategy.
  - This removes the requirement of the FinalizationRegistry browser feature which is not available prior to chrome 84 / WPE 2.38. 
  - The prior behavior is still available by setting a new experimental renderer setting to `true`:
    - `experimental_FinalizationRegistryTextureUsageTracker`
- Added `texture-memory-stress` test with instructions on how to perform.
- Example tests may now export a `customSettings` function which returns any renderer settings overrides that the test requires to run. These settings are applied on top of all of the defaults specified by the example test launcher.
- Improve type checking for EffectDesc
  - `type` now determines the specific properties types checked.
- **Breaking changes:**
  - Refactor ShaderDesc / TextureDesc related areas including change their names to ShaderRef / TextureRef.
    - RendererMain.makeTexture renamed to RendererMain.createTexture
    - RendererMain.makeShader renamed to RendererMain.createShader
    - TextureDesc type renamed to TextureRef type
    - ShaderDesc type renamed to ShaderRef type

### DIFF
--- a/examples/common/Character.ts
+++ b/examples/common/Character.ts
@@ -18,10 +18,11 @@
  */
 
 import type {
+  SpecificTextureRef,
   INode,
   INodeWritableProps,
   RendererMain,
-  TextureDesc,
+  TextureRef,
 } from '@lightningjs/renderer';
 import { assertTruthy } from '@lightningjs/renderer/utils';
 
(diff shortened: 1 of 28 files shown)


frank-weindel:
@erikhaandrikman Could I get your thoughts on this approach?

wouterlucas:
Would there ever be a case where I'd want to run ThreadX without FinalizationRegistry?

jfboeve (approved):

frank-weindel:
> Would there ever be a case where I'd want to run ThreadX without FinalizationRegistry?

FinalizationRegistry for texture memory isn't really coupled to ThreadX at all. It's just one strategy for proactively cleaning up textures that we've cached and loaded on the GPU that we are pretty confident won't be needed again for awhile. In the case of the FinalizationRegistry strategy we know 100% that a texture is no longer referenced anywhere in the currently running JavaScript at all. The new Manual Count + Interval strategy employs a time heuristic that aims to predict when that is the case, but with some chances of being wrong. There are pros/cons to both approaches. GC and the firing of the FinalizationRegistry events for instance isn't guaranteed in any particular timeframe by the browser. It could come super late, for instance, to a point where it's not being proactive enough to prevent memory pressure overload. The Manual Count strategy is guaranteed to run and clean up unused textures regularly.

PR #37, opened 2023-10-05
fixed mixing gradient colors with maskcolor

jfboeve:
Chris found an issue with gradients that contained alpha lower than 1. This PR resolves that issue.

### DIFF
--- a/examples/tests/dynamic-shader.ts
+++ b/examples/tests/dynamic-shader.ts
@@ -154,20 +154,23 @@ export default async function ({ renderer }: ExampleSettings) {
     y: 100,
     width: 250,
     height: 500,
-    color: 0x00000000,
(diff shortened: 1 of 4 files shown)


examples/tests/dynamic-shader.ts
```diff
@@ -154,20 +154,23 @@ export default async function ({ renderer }: ExampleSettings) {
     y: 100,
     width: 250,
     height: 500,
-    color: 0x00000000,
+    color: 0xff0000ff,
```
frank-weindel, line 157: This PR totally LGTM! When you get some time could you try to work in the new row pagination system I created for `clipping` test? It will let you display many more examples (since it paginates) and let you describe what you're testing (as row headers).

Issue #26, opened 2023-10-02
Support TV Devices


chiefcll:
Let's get the renderer running on TV devices. Currently they don't support offscreen canvas and FinalizationRegistry. Let's put those two things behind a feature flag or use detection and provide an alternative.

wouterlucas:
Original "make it work" PR: https://github.com/lightning-js/renderer/pull/7

@frank-weindel is working on a long term PR to be able to only use FinalizationRegistry / OffscreenCanvas when the ThreadX driver is selected (as both are needed in that case).

iamtabrezkhan:
Can I work on this? @wouterlucas

wouterlucas:
of course @iamtabrezkhan!

frank-weindel:
#46 should be enough to close this out for at least production Comcast devices. @iamtabrezkhan could you confirm from your perspective?

wouterlucas:
Confirmed on discord, tested on XG2v2

PR #41, opened 2023-10-06
FIX: Offscreen Canvas context is not available

iamtabrezkhan:
This PR aims to solve [this issue](https://github.com/lightning-js/renderer/issues/26#issuecomment-1750438486).
Solution: If the `OffscreenCanvas` is available but it's `context` is not, the code falls back to using `HTMLCanvasElement` and it's context.


frank-weindel (approved):
LGTM! Thanks for the contribution! I've added some comments to clarify why this is needed.

PR #42, opened 2023-10-06
Fix looping animations with `stopMethod: 'reverse'`

erikhaandrikman:
This fixes looping an animation that has `stopMethod: 'reverse'`

### DIFF
--- a/examples/tests/animation.ts
+++ b/examples/tests/animation.ts
@@ -17,30 +17,17 @@
  * limitations under the License.
  */
 
+import type { IAnimationController } from '@lightningjs/renderer';
+
 import type { ExampleSettings } from '../common/ExampleSettings.js';
+interface AnimationExampleSettings {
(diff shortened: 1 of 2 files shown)


examples/tests/animation.ts
```diff
@@ -91,25 +86,65 @@ export default async function ({ renderer }: ExampleSettings) {
...
-          duration: 3400,
-          loop: true,
-          easing: easings[i],
-        },
-      )
-      .start();
+  let animationIndex = 0;
+  let currentAnimation: IAnimationController;
+
+  const animationSettings: Partial<AnimationExampleSettings> = {
```
frank-weindel, line 95: The PR's fix LGTM. But I think you might've forgot to pass the easing to the animationSettings structure in the test example?

frank-weindel (approved):

PR #51, opened 2023-10-11
Refactor Effects + changes

jfboeve:
- Refactored the Effects to the ShaderManager, this way a developer can add their own custom effects.
- Added more docs on the FadeOutEffect and RadiusEffect.
- *breaking* Angle property is now in Radians not in Degrees. The reason for this change is that we only use radians in the core, and it saves a calculation for each pixel from degrees to radians. The dynamic-shader example has a example function that calculates degrees to radians (PI / 180 * degrees)

### DIFF
--- a/src/core/CoreShaderManager.ts
+++ b/src/core/CoreShaderManager.ts
@@ -22,9 +22,24 @@ import type { CoreShader } from './renderers/CoreShader.js';
 
 import { DefaultShader } from './renderers/webgl/shaders/DefaultShader.js';
 import { DefaultShaderBatched } from './renderers/webgl/shaders/DefaultShaderBatched.js';
-import { DynamicShader } from './renderers/webgl/shaders/DynamicShader.js';
+import {
+  DynamicShader,
+  type DynamicShaderProps,
+} from './renderers/webgl/shaders/DynamicShader.js';
 import { RoundedRectangle } from './renderers/webgl/shaders/RoundedRectangle.js';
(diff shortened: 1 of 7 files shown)


src/core/renderers/webgl/shaders/effects/RadiusEffect.ts
```diff
@@ -29,6 +29,20 @@ export interface RadiusEffectProps extends DefaultEffectProps {
   /**
    * Corner radius, in pixels, to cut out of the corners
    *
+   * You can input an array with a length of up to four or a number.
```
frank-weindel, line 32: Just for the future, it's better to place all these longer pieces of explanation under a `@remarks` header.

src/core/CoreShaderManager.ts
```diff
@@ -62,6 +103,13 @@ export class CoreShaderManager {
     this.shConstructors[shType] = shClass;
   }
 
+  registerEffectType<Type extends keyof EffectMap>(
```
frank-weindel, line 106: Is the `ShaderEffect` base class exported via the Core API export (exports/core-api.ts)? I don't think it is at first glance. Users will need that as well in order to create custom effects.

src/core/renderers/webgl/shaders/effects/RadiusEffect.ts
jfboeve: done

src/core/CoreShaderManager.ts
jfboeve: Done

PR #61, opened 2023-10-24
Add Backstop testing

chiefcll:


.github/workflows/tests.yml
```diff
@@ -0,0 +1,63 @@
...
+    steps:
+      - name: Use Node.js 16.x
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16.x
+
+      - name: Checkout renderer
+        uses: actions/checkout@v3
+
+      - name: Checkout threadx
```
uguraslan, line 18: This step and the following threadx building step are not necessary anymore. I think both steps should be removed.

backstop.json
```diff
@@ -0,0 +1,29 @@
+{
+  "id": "backstop_default",
+  "viewports": [
+    {
+      "label": "tv",
+      "width": 1920,
```
frank-weindel, line 6: Good stuff. Can we just test 720p without the extra white frame around the window?

erikhaandrikman: Yes, will change the test viewport to `720p`

src/core/animations/AnimationManager.ts
```diff
@@ -19,6 +19,14 @@
 
 import { CoreAnimation } from './CoreAnimation.js';
 
+// This temporary implementation disables frame-by-frame animation
+// using requestAnimationFrame and locks it to a fixed position in the animation.
+// It emulates the delta time between two frames, a value typically calculated
+// as the difference between the current and previous frame.
+// A more robust and sustainable solution will be needed in the future.
+const urlParams = new URLSearchParams(window.location.search);
+const delta = parseInt(urlParams.get('delta') as string);
```
frank-weindel, line 28: We def can't have DOM stuff leaking into the Core space. I think all the delta emulation should be moved up into Stage.drawFrame(). The AnimationManager shouldn't need to be any the wiser about where its delta comes from.

frank-weindel: I was thinking methods accessible from RendererMain that allow the renderer to be paused, and once it's paused it can be advanced manually in N millisecond increments.

erikhaandrikman: Fully agree, the existing code is in place to facilitate rapid testing and enhance the discussion regarding precise animation testing. Feel free to make any adjustments as needed

wouterlucas:
Suggestion is to take the animation bits out, roll this in. Seperate PR for the animations

erikhaandrikman:
This PR is being closed as it has been replaced by the following: https://github.com/lightning-js/renderer/pull/78

PR #67, opened 2023-10-26
Add some more test examples

chiefcll:
Wanted to demo how to do a shader on an image and add some tests for them as well.

Clipping with border radius doesn't currently work. I assume this needs RTT? Do we need an issue to support this?

Added image for different gradients. Not sure which ones should and shouldnt work - this is what it looks like now:
image

### DIFF
--- a/examples/tests/background-image-shader.ts
+++ b/examples/tests/background-image-shader.ts
@@ -1,0 +1,51 @@
+/*
+ * If not stated otherwise in this file or this component's LICENSE file the
+ * following copyright and licenses apply:
+ *
+ * Copyright 2023 Comcast Cable Communications Management, LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the License);
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { ExampleSettings } from '../common/ExampleSettings.js';
+
+export default async function ({ renderer }: ExampleSettings) {
+  const t0 = renderer.createNode({
+    x: 0,
+    y: 0,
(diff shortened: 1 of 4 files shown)


chiefcll:
Clipping
image

examples/tests/background-image-shader.ts
```diff
@@ -0,0 +1,51 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { ExampleSettings } from '../common/ExampleSettings.js';
+
+export default async function ({ renderer }: ExampleSettings) {
+  const t0 = renderer.createNode({
```
frank-weindel, line 23: To me this just seems like a test for the DynamicShader with images. In the interest of reducing duplicative tests, do you think we can just rely on you created in `image-dynamic-shader`?

examples/tests/clipping-border-radius.ts
```diff
@@ -0,0 +1,533 @@
...
+
+const SQUARE_SIZE = 200;
+const PADDING = 20;
+
+export default async function ({ testName, renderer }: ExampleSettings) {
+  const pageContainer = new PageContainer(renderer, {
+    width: renderer.settings.appWidth,
+    height: renderer.settings.appHeight,
+    parent: renderer.root,
+    title: 'Clipping Tests',
```
frank-weindel, line 32: You're right that this is not possible without RTT. The `clipping` property, just like in Lightning 2, only supports clipping directly around the non-rotated rectangular bounds of a node. Once we implement it, the "clipping" ability of RTT+Shaders will easily be testable in a single test example for RTT. So I think you should delete this one in anticipation for that.

examples/tests/image-dynamic-shader.ts
```diff
@@ -0,0 +1,243 @@
...
+
+import type { ExampleSettings } from '../common/ExampleSettings.js';
+import robot from '../assets/robot/robot.png';
+
+export default async function ({ renderer }: ExampleSettings) {
+  const degToRad = (deg: number) => {
+    return (Math.PI / 180) * deg;
+  };
+
+  const t1 = renderer.createNode({
```
frank-weindel, line 28: Maybe its subtle but I noticed that two of the test cases here don't seem to be displaying an image, perhaps you can work an image into them as well so they seem less duplicated from the standard dynamic-shader example?

chiefcll: Yea - some of this was for @jfboeve to be aware. I think it's important we have tests which cover DynamicShader on an image src as lot of users will want to do that.

Issue #34, opened 2023-10-03
Add width / height to ImageTexture


chiefcll:
If I load an image without a width and height - it doesn't display.

However, on txLoaded I get the width and height back and if I then assign it to the lightning node, it will render. I think the width and height should be assigned automatically.

frank-weindel:
Thanks @pecoram for your PR (#68)! Before we proceed any further I think we should have a discussion on the how best to implement the feature request to make sure we cover all the edge cases and keep the implementation as simple as possible.

I think it's first helpful to break the request down into the related use cases:
- **Base Node Use Cases**
  - User wants to be in complete control of the node's width/height.
    - This could be because the user wants a specific size to stretch the texture to or for child clipping purposes.
  - User wants node to be automatically resized to whatever the latest loaded texture is.
- **Text Node Use Cases**
  - User wants to be in complete control of the node's width/height (`contain: 'both'`)
  - User wants to set a specific width while the height is automatically resized (`contain: 'width'`)
  - User wants node to be automatically resized to whatever the latest loaded text is (`contain: 'none'`)

For all the use cases (feel free to miss any I'm missing), we need some deterministic/intuitive way for the user to either opt-in to or opt-out of autosizing behavior. For Text Nodes, we could probably just rely on the `contain` property for that. For Base Nodes, we need some way for the user to decide. I'm not a fan of using `0` width/height for this because that should be a perfectly valid value for the width/height that simply results in nothing seen. The width/height properties cannot be `null`/`undefined` so the user couldn't use those to decide either. I think overloading an intuitively and perfectly valid value (`0`) with an extra meaning in this case could lead to issues and confusion down the road. Imagine a user wants to start a Node out at 0x0 and after it's texture loads it wants to animate it to the appropriate width/height.

## `autosize` Property Proposal

For this, I propose we add a new boolean property `autosize` that is `false` by default. When it's `true` the autosizing behavior discussed here will occur. When it is set to `false` the current behavior of not changing the Node's size (even for Text Nodes, since they will inherit this property) should occur. To make the likely more common case where the user wants the node to be autosized with an image, when the user sets the `src` property, the `autosize` property will be set to true automatically as a side effect. This will require users to be explicit about setting `autosize` to true if they want to set the `texture` property manually (instead of using `src`), which I think is special enough of a use case that it won't cause much grief/confusion for the developer.

Feedback welcomed!

frank-weindel:
@michielvandergeest @erikhaandrikman for awareness

[...]

PR #68, opened 2023-10-26
feat: Add width / height to ImageTexture

pecoram:
should to fix https://github.com/lightning-js/renderer/issues/34

- without width and height the image width, height are set
- with width, height is set with aspect ratio
- with height, width is set with aspect ratio


frank-weindel (changes requested):
Thanks for the PR @pecoram! Please discuss with us in #34. I'm not a fan of tying this directly to 0 values and we do need to consider Text in how this functions.

frank-weindel:
Labeling this as a breaking change as it's possible downstream frameworks/apps will behave differently without changes when this feature is implemented.

src/core/CoreNode.ts
```diff
@@ -164,7 +166,34 @@ export class CoreNode extends EventEmitter implements ICoreNode {
     this.props.textureOptions = null;
   }
 
+  autosizeNode(dimensions: Dimensions) {
+    if (this.autosize && dimensions.width && dimensions.height) {
+      if (
+        (isNaN(this.width) && isNaN(this.height)) ||
+        (this.width === 0 && this.height === 0)
```
frank-weindel, line 173: We shouldn't try to handle aspect ratios. If the `autosize` property is true, it should just set the width/height to the width and height of the texture.

src/core/CoreNode.ts
```diff
@@ -164,7 +166,34 @@ export class CoreNode extends EventEmitter implements ICoreNode {
...
+          this.width = this.height * imageAspectRatio;
+        } else if (isNaN(this.height) || this.height === 0) {
+          this.height = this.width / imageAspectRatio;
+        }
+      }
+    }
+  }
+
   private onTextureLoaded: TextureLoadedEventHandler = (target, dimensions) => {
+    if (target.props.src) {
```
frank-weindel, line 193: Keep in mind that the `src` property does not exist at all in the Core API Space. It's implementation only exists in the Main API Space, split between MainOnlyNode.ts and ThreadXMainOnlyNode.ts.

https://github.com/lightning-js/renderer/blob/6e9dc53d0dd58a7c949e31d5d4db0770e8b01f5e/src/render-drivers/main/MainOnlyNode.ts#L347C12-L363

https://github.com/lightning-js/renderer/blob/6e9dc53d0dd58a7c949e31d5d4db0770e8b01f5e/src/render-drivers/threadx/ThreadXMainNode.ts#L109-L125

chiefcll (commented):
👍

src/main-api/INode.ts
```diff
@@ -98,7 +97,7 @@ export interface INodeWritableProps {
    * TODO: Add support for non-rectangular clipping either automatically or
    * via Render-To-Texture.
    *
-   * @default `false`
+   * @default `null`
```
frank-weindel, line 100: I think this is unintended

src/main-api/INode.ts
```diff
@@ -77,11 +77,10 @@ export interface INodeWritableProps {
    *
    * @remarks
    * The autosize to true let the image node to be resized with the original size of the image.
-   * Combined with one of width and height the other size will be set manteining the aspect ratio
    *
    * @default `false`
    */
-  autosize: boolean;
+  autosize: boolean | null;
```
frank-weindel, line 83: `null` will not work as a value on the ThreadX implementation. For now, just make setting `src` (in both MainOnlyNode and ThreadXMainNode) also set `autosize` to true unconditionally

frank-weindel, line 83: `null` will not work as a value on the ThreadX implementation. For now, just make setting `src` (in both MainOnlyNode and ThreadXMainNode) also set `autosize` to true unconditionally

src/core/CoreNode.ts
```diff
@@ -168,31 +168,13 @@ export class CoreNode extends EventEmitter implements ICoreNode {
 
   autosizeNode(dimensions: Dimensions) {
     if (this.autosize && dimensions.width && dimensions.height) {
```
frank-weindel, line 170: I don't think we need to condition this on width/height being truthy. If the dimensions are truly somehow 0x0 then autosizing should still occur

src/main-api/INode.ts
frank-weindel: I understand what you're trying to do here. But at the moment there's a limitation in how that can be done. For now we'll need to rely on the order of setting the properties to allow the user to set `autosize` to false if they also set `src`. This is something we can improve on in the future.

PR #80, opened 2023-11-07
Implement Text Node Parent Alpha

### LINKED ISSUE #79: Text Node Parent Alpha
Currently the combined alpha of the descendent nodes of a TextNode are not applied to the Text Node itself. This means text may still appear fully opaque even if its parent (or further descedents) has an alpha less than 1.0

The ultimate alpha value that text is rendered should be the combined multiplied alpha of its decedents multiplied by its own alpha value (including it's own color alpha). The color alphas of decedents should not be considered in this calculation as they are used solely for rendering that particular node's texture or rectangle color.

frank-weindel:
Resolves #79

### DIFF
--- a/examples/common/constructTestRow.ts
+++ b/examples/common/constructTestRow.ts
@@ -20,14 +20,26 @@
 import type { INode, RendererMain } from '../../dist/exports/main-api.js';
 import { waitForTextDimensions } from './utils.js';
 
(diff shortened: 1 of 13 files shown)


chiefcll:
Looks great! Exciting

chiefcll (approved):

PR #72, opened 2023-10-31
Standardize the loaded event of text and texture

Drulokia:
Standardized event for text and texture loaded.
I guess the events for text and texture failed event can be added too.

PR #81, opened 2023-11-07
Fix up issues with Text / Texture loaded/failed event refactor

frank-weindel:
There were some issues that got merged in from #72 that caused some examples to break. I've fixed the issues and also refactored these types a bit more to reduce some ambiguity that had been present even before #72.


wouterlucas (approved):
LGTM :)

PR #84, opened 2023-11-13
Fix issue that caused text moved from offscreen to not render properly

### LINKED ISSUE #50: Multiline text positioned outside the viewport only renders first line
When a text node is positioned (at creation) outside of the visible viewport (i.e outside the 1920x1080 , with the default settings) it will only render the first line. Until a rerender of that text node is triggered (by changing the color or the fontsize for example) it will render the full multiline text.

Can be reproduced with the following example:

```js
import type { ExampleSettings } from '../common/ExampleSettings.js';

export default async function ({ renderer }: ExampleSettings) {

  const parent = renderer.createNode({
    x: 100,
    y: 100,
    color: 0x00000000,
    parent: renderer.root,
  });


  const texts = [
    'Lorem ipsum dolor sit amet, consectetur adipiscing elit.',
    'Sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.',
    'Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.',
    'Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.',
    'Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.',
    'Nulla ac libero nec justo blandit fringilla.',
    'Vestibulum auctor nunc at augue rhoncus, id vestibulum nulla aliquet.',
    'Maecenas in sem euismod, vehicula leo ut, commodo quam.',
    'Curabitur eu nunc sit amet urna euismod consequat.',
    'Pellentesque habitant morbi tristique senectus et netus et malesuada fames ac turpis egestas.'
  ]

  const nodes = []

  texts.forEach((text, index) => {

    const node = renderer.createTextNode({
      y: index * 250,
      //y: index * 100, // show all texts completely (because all are 'in viewport")
      width: 500,
      color: 0xff0000ff,
      parent,
      fontFamily: 'Ubuntu',
      fontSize: 40,
      contain: 'width',
      text
    });

    nodes.push(node)

  })

  const interval = setInterval(() => {
    parent.y -= 100
    if(parent.y <= -1600) clearInterval(interval)
  }, 500)

  setTimeout(() => {
    nodes[nodes.length - 1].color = 0xff00ffff
    setTimeout(() => {
      nodes[nodes.length - 2].color = 0xffffffff
    }, 2000)
    setTimeout(() => {
      nodes[nodes.length - 3].fontSize = 50
    }, 3000)
  }, 10000)


}
```

frank-weindel:
Also included are some performance optimizations on how text renderer state gets updated.

Resolves #50

### DIFF
--- a/examples/tests/text-offscreen-move.ts
+++ b/examples/tests/text-offscreen-move.ts
@@ -36,6 +36,15 @@ const commonTextProps = {
   fontSize: 50,
 } satisfies Partial<ITextNodeWritableProps>;
 
+/**
+ * This test is to ensure that text that starts offscreen and moves onscreen
+ * is rendered correctly.
+ *
+ * This test was designed around the following bug report:
+ * https://github.com/lightning-js/renderer/issues/50
+ *
+ * @param param0
+ */
 export default async function ({ renderer, testName }: ExampleSettings) {
   const pageContainer = new PageContainer(renderer, {
     width: renderer.settings.appWidth,
@@ -51,6 +60,8 @@ export default async function ({ renderer, testName }: ExampleSettings) {
   pageContainer.pushPage(createTestCase(renderer, 'canvas', 'none'));
   pageContainer.pushPage(createTestCase(renderer, 'canvas', 'width'));
   pageContainer.pushPage(createTestCase(renderer, 'canvas', 'both'));
+
+  await delay(200);
   pageContainer.finalizePages();
 
   pageContainer.bindWindowKeys();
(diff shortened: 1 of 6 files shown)


michielvandergeest (approved):
LGTM! Tested locally and these changes successfully seem to fix the issue

PR #91, opened 2023-11-23
Render loop optimizations

wouterlucas:
We've spent a few rounds of doing render loop optimizations and discovered there are several improvements that can be made. Consider this round # 1 of our performance hunt. 

Current list of changes:
- [x] Pause RAF when there are no updates
- [x] Do not render if there are no tree updates
- [x] Animation loop code optimizations
- [x] Change node sorting to presorting
- [x] Disable Batched Shader for now, due to branching issues
- [ ] Excessive texture rebinds per draw 

**Benchmarking**
We've used two benchmarking methods:
- Blits 1080p stress test of a continuous animation with 103 images
- Simple 720p animation on L2 and L3

Observations, all tested on a RPI3 with WPEWebKit 2.42:
Simple animation on L2 @ 720p rendered at ~59 FPS
Simple animation on L3 @ 720p rendered at 33 FPS at the start of this exercise
Simple animation on L3 @ 720p renders at 47 FPS with the above changes, excluding texture rebinds
Simple animation on L3 @ 720p renders at ~59 FPS with a texture rebind hack (TODO implementation)

Blits stress test @ 1080p renders at 12 FPS prior to these changes
Blits stress test @ 1080p renders at 27 FPS with the above changes
Blits stress test @ 1080p renders at ~47 FPS with a texture rebind hack (TODO implementation)

@erikhaandrikman for all the support & hunting :love_you_gesture: 

Todo:
- [] Come up with a way to handle texture rebinds
- [] Run visual regression testing


src/main-api/RendererMain.ts
```diff
@@ -199,6 +199,13 @@ export interface RendererMainSettings {
    */
   experimental_FinalizationRegistryTextureUsageTracker?: boolean;
 
+  /**
+   * Pause the render loop if there are no scene updates
+   *
+   * @defaultValue `true`
+   */
+  pauseRaf?: boolean;
```
frank-weindel, line 207: Do we want to ship this as an option vs just making it the only behavior?

src/core/renderers/webgl/shaders/DefaultShader.ts
```diff
@@ -37,6 +37,10 @@ export class DefaultShader extends WebGlCoreShader {
 
   override bindTextures(textures: WebGlCoreCtxTexture[]) {
     const { gl } = this;
+    if (gl.getParameter(gl.TEXTURE_BINDING_2D) === textures[0]!.ctxTexture) {
```
frank-weindel, line 40: I could work on this as a more universal fix since this only applies to the DefaultShader vs SDF text for instance. Let me know.

src/main-api/RendererMain.ts
wouterlucas: Either way works for me, I copied the configurability from L2. There we added it as an opt-in because it was different then what earlier releases where doing. Though with the latest WebKit devices we seem to almost always require to pause raf to alleviate the system resources especially on certain compositors. 

If you prefer I can drop the configuration part for now, we can always add it back in case its an issue somewhere.

src/core/renderers/webgl/shaders/DefaultShader.ts
wouterlucas: yeah - consider this a stop-gap solution, as it does help a bit in overall FPS. 
I'd suggest we move this in, so you can work on a more universal fix on your own pace and then replace it down the road?

PR #95, opened 2023-12-04
Implement `fpsUpdate` event and `fpsUpdateInterval` setting on RendererMain

frank-weindel:
Per second FPS logging can be enabled on Example Tests via the new `fps` URL param


wouterlucas (approved):
LGTM - I would even argue to have the FPS turned on by default on all the examples. But that might be me being lazy =)

Issue #86, opened 2023-11-20
Support of line height for multi line text


suresh-gangumalla:
When rendering multil-line text, came across use of the line height for proper spacing. Currently, renderer does not seem to  provide a way to set the line height.

PR #98, opened 2023-12-05
fix: added missing text props

pecoram:
fixes https://github.com/lightning-js/renderer/issues/86

Added the following props:

- lineHeight
- maxLines
- verticalAlign
- textBaseline
- overflowSuffix

These props are not yet compatible with the SDF renderer (https://github.com/lightning-js/renderer/issues/107)

More Props not added:
- textOverflow

- [x] visual regression tests @frank-weindel


src/core/text-rendering/renderers/TextRenderer.ts
```diff
@@ -244,6 +244,10 @@ export interface TrProps extends TrFontProps {
    * @default 0
    */
   letterSpacing: number;
+  lineHeight?: number | null;
+  maxLines?: number | undefined;
+  maxLinesSuffix?: string | undefined;
+  textOverflow?: 'ellipsis' | 'clip' | undefined;
```
frank-weindel, line 250: These shouldn't be optional

src/render-drivers/main/MainOnlyTextNode.ts
```diff
@@ -197,6 +200,46 @@ export class MainOnlyTextNode extends MainOnlyNode implements ITextNode {
     this.coreNode.letterSpacing = value;
   }
 
+  get lineHeight(): ITextNode['lineHeight'] {
```
frank-weindel, line 203: ThreadX support also needed

src/core/text-rendering/renderers/TextRenderer.ts
```diff
@@ -244,6 +244,10 @@ export interface TrProps extends TrFontProps {
    * @default 0
    */
   letterSpacing: number;
+  lineHeight?: number | null;
+  maxLines?: number | undefined;
```
frank-weindel, line 248: Please implement only text `lineHeight` in this PR. `maxLines` is already implemented by setting the height to `lineHeight * numLines` and contain to `both`. There is a plan to handle `maxLinesSuffix` and `textOverflow` differently.

pecoram: Are you sure that `maxLines` is not needed? Consider this case:
I want that a node adapts his height based on the text but no more then `lineHeight * numLinesà. So the height can go from `lineHeight` to `lineHeight * maxLines `

src/core/text-rendering/renderers/CanvasTextRenderer.ts
```diff
@@ -317,6 +325,8 @@ export class CanvasTextRenderer extends TextRenderer<CanvasTextRendererState> {
         wordWrapWidth:
           state.props.contain === 'none' ? undefined : state.props.width,
         letterSpacing: state.props.letterSpacing,
+        lineHeight: state.props.lineHeight,
+        maxLines: state.props.maxLines,
```
frank-weindel, line 329: For a `maxLines` implementation for the CanvasTextRenderer, I'd like to base it off of the `contain` === 'both' mode like so (semi-pseudocode):
```
{
  //...
  maxLines: contain === 'both' ? Math.floor(height / lineHeight) : undefined
}
```
Where `lineHeight` is the set lineHeight property, if provided, or the set `fontSize` property.

SDF Text Renderer already implements this capability in basically this way.

pecoram: this is ok but it is not suitable for dynamic height text nodes. See my example image.

frank-weindel: @pecoram Where can I find your images?

pecoram: https://github.com/lightning-js/renderer/pull/98#discussion_r1424073231

frank-weindel: Got it. Yeah that seems like a valid use case to me. We'll just want to make sure that the contained height and maxLines work well together. For instance, if maxLines is 99, but the contained height can only fit 5 lines then only 5 lines are rendered before the text is truncated.

pecoram: image

```
height: 36
lineHeight: 36
maxLines: 99
contains: both
```

wouterlucas:
Really cool @pecoram 😎 thanks!

I dont want to be that guy, but doing it anyway, any chance we can get an accompanying examples test that demonstrates/tests this?

Would be nice to have this included in future visual regression tests.

pecoram:
I added some visual regression tests

src/core/text-rendering/renderers/LightningTextTextureRenderer.ts
```diff
@@ -212,7 +212,7 @@ export class LightningTextTextureRenderer {
       this._settings.offsetY === null
         ? null
         : this._settings.offsetY * precision;
-    let lineHeight = (this._settings.lineHeight || fontSize) * precision;
+    let lineHeight = (this._settings.lineHeight || fontSize * 1.4) * precision;
```
frank-weindel, line 215: Where does this `1.4` value come from? One of the reason that many of the existing tests needed to be re-done.

src/core/text-rendering/renderers/LightningTextTextureRenderer.ts
```diff
@@ -673,10 +673,10 @@ export class LightningTextTextureRenderer {
       lineHeight: null,
       textBaseline: 'alphabetic',
       textAlign: 'left',
-      verticalAlign: 'top',
+      verticalAlign: 'middle',
```
frank-weindel, line 676: We should default to verticalAlign `'top'`

frank-weindel:
Everything looks good beside my comments. I went ahead and pushed some new commits that resolve my issues. So I'm ready to merge this now 😄 

Thank you @pecoram !

PR #103, opened 2023-12-07
Perf: Implement flag driven update dependency graph

frank-weindel:
dependency-graph

Updates to cached calculated data within a CoreNode is now updated according to the above dependency graph.

New optimizations:
- Positioning, sizing and scaling properties now set an `updateType` (formerly `recalculationType`) flag to indicate an update is needed to their corresponding matrix.
- Children are not visited unless their `updateType` is flagged with `Children`
- Any updates to a CoreNode efficiently flags all ancestors up to the root with `Children`
- Children are not visited if their own `updateType` is 0.
- World Alpha is now update cached.
- Premultiplied Alpha-merged colors are now update cached.
- Calculated Z-Index is now update cached.

Other changes:
- TypeDocs output folder renamed from `docs` to `typedocs`.
- New `docs` folder checked in containing manually written documentation.
- Added documentation about the Node Dependency Graph

On a RPI test, this took a resolution=1080, multiplier=1, overlay=false `stress` test from about 40-41 fps to 42 fps.


frank-weindel:
There's a bug with alpha I need to look into it seems

PR #110, opened 2023-12-20
Context Spy

frank-weindel:
When enabled by the `enableContextSpy` renderer option, the canvas context object (WebGL in the current case) will be instrumented with a spy mechanism that counts the number of calls to each context function. The number of calls to each function during the previous FPS interval will be included in the new `contextSpyData` key of the 'fpsUpdate' event.

The Context Spy can be enabled in Example Tests by the `contextSpy` boolean URL param. When enabled, a statistical calculation of each context function call will be reported after every FPS statistics summary. Because of this, the option requires that the `fps` URL param is also set to true, otherwise only a degradation of performance will occur without any output of the Context Spy data.

Example output:
```
---------------------------------
Average FPS: 4.4
Median FPS: 4
P01 FPS: 4
P05 FPS: 4
P25 FPS: 4
Std Dev FPS: 0.48989794855663565
Num samples: 5
---------------------------------
median(disable) / median(fps): 4
median(clear) / median(fps): 1
median(bindBuffer) / median(fps): 1
median(bufferData) / median(fps): 1
median(getParameter) / median(fps): 3
median(activeTexture) / median(fps): 3
median(bindTexture) / median(fps): 3
median(uniform2fv) / median(fps): 3
median(uniform1f) / median(fps): 3
median(enable) / median(fps): 3
median(blendFunc) / median(fps): 3
median(drawElements) / median(fps): 3
median(totalCalls) / median(fps): 28
---------------------------------
````

**BREAKING CHANGE**

This is a breaking change because the 'fpsUpdate' event previously only used a `number` as its payload type, whereas now it is an object with the keys `fps` and `contextSpyData`. Apps/frameworks that listen for the 'fpsUpdate' event will need to update their handlers.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -50,6 +50,16 @@ pnpm watch
   - Whether or not to log the latest FPS sample to the console every 1 second.
   - After skipping the first 10 samples, every 100 samples after that will result
     in a statistics summary printed to the console.
+- `contextSpy` (boolean, default: "false")
+  - Whether or not to turn on the context spy and reporting
+  - The context spy intercepts all calls to the (WebGL) context and reports
+    how many calls to each function occurred during the last FPS sampling period
+    (1 second for these tests).
+  - Statistical results of every context call will be reported along with the
+    FPS statistics summary.
+  - `fps` must be enabled in order to see any reporting.
+  - Enabling the context spy has a serious impact on performance so only use it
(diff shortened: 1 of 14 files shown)


wouterlucas (approved):
Neat! This would be great base for a more advanced debugger of some kind, getting these calls/statistics is very useful.

PR #113, opened 2023-12-20
Fix issue that caused animations not to finish all the way

frank-weindel:
This fixes the reported issue where animations do not always progress all the way to their destination before stopping.


michielvandergeest (approved):
Tested with Blits transitions example and this seems to fix the issue I experience earlier. Transitions now finish correctly with their final value.

PR #115, opened 2023-12-20
Fix potential endValue issue if timingFunction is incorrectly implemented.

wouterlucas:
In the original code, currentValue can exceed endValue if the timingFunction returns a value greater than 1 when passed a progress of 0.99.

To ensure that currentValue never exceeds endValue, regardless of the output of timingFunction, using Math.min to limit currentValue to endValue.

### DIFF
--- a/src/core/animations/CoreAnimation.ts
+++ b/src/core/animations/CoreAnimation.ts
@@ -93,8 +93,12 @@ export class CoreAnimation extends EventEmitter {
     }
   }
 
-  applyEasing(p: number, s: number, e: number): number {
-    return (this.timingFunction(p) || p) * (e - s) + s;
+  applyEasing(progress: number, startValue: number, endValue: number): number {
+    const easedProgress = this.timingFunction(progress) || progress;
+    const currentValue = easedProgress * (endValue - startValue) + startValue;
+
+    // Ensure currentValue doesn't exceed endValue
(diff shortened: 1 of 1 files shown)


frank-weindel:
I don't think this is a bug? If an easing function dictates that the animated property overshoots its end value before reaching its end value at `progress = 1`, that should be allowed.

wouterlucas:
Does it? Would an easing function be allowed to reach (even at progress = 1) state where the animated property overshoots the desired end state? 

It would create a similar case right, where you ask a property to animate to a `x` position of `350` and the easing function could place it at `355` respectively.

frank-weindel:
Not at progress = 1, but, for example, at progress 0.95 it certainly could. If the easing function at progress = 1 does not map to the end value, that should be considered a bug in the easing function. Along the same lines, theoretically progress = 0 could map to something other than the start value if the easing function was not implemented correctly.

wouterlucas:
Right - well I was hunting for potential places that could result into overshoot. Assuming we would always trust the easing functions OR this would be expected this isn't needed.

Feel free to close it if we don't want to check it here. 

Also looking back, likely a ternary check would be slightly faster for the JIT over Math.min 😄

frank-weindel:
Cool. We can re-open down the road if needed.

PR #116, opened 2023-12-28
Update grayscale effect to fall in-line with L2 shader version

jfboeve:
The old Grayscale shader, had an `amount` prop which was missing in this version of the grayscale effect. This PR fixes the grayscale to fall in-line with what was expected of the grayscale shader in Lightning 2


suresh-gangumalla:
Verified the PR and its working as expected.

michielvandergeest (approved):

frank-weindel (approved):

PR #119, opened 2024-01-09
BorderEffect alpha fix

jfboeve:
Border alpha was not properly calculated so I fixed the formula


michielvandergeest (approved):
great! this works as expected now.

frank-weindel (approved):

PR #121, opened 2024-01-17
CoreNode: Set renderable to false if worldAlpha is 0

frank-weindel:


chiefcll (approved):

wouterlucas (approved):
LGTM

chiefcll:
@frank-weindel - can you remove the duplicate check:
https://github.com/lightning-js/renderer/blob/main/src/core/Stage.ts#L263

frank-weindel:
@chiefcll Let's just leave the existing working implementation. I don't think this implementation as is will work if a node has an image/texture on it since `isRenderable` will be set to true. I'd like to revisit `isRenderable` as I think we can reduce how many times we call it all together and perhaps solve the reason we need the `texture` check at the top.

The `texture` check at the top exists because we want to allow a 0x0 node to have a texture assigned to it and then have that texture load and produce the loaded event which a node can resize itself with. Without that check at the top 0x0 nodes don't even attempt to load the texture because the node is never rendered to trigger the texture load.

PR #128, opened 2024-01-19
Feat/image worker

chiefcll:
Add a web worker to load images.

@frank-weindel - Ideally the numWorkers should be passed into the stage and passed through. I wasn't sure the cleanest way of doing that part.

### DIFF
--- a/src/core/lib/ImageWorker.ts
+++ b/src/core/lib/ImageWorker.ts
@@ -34,7 +34,9 @@ export class ImageWorkerManager {
       this.workers.forEach((worker) => {
         worker.onmessage = this.handleMessage.bind(this);
       });
+      console.log('Image Workers Enabled numWorkers:', numImageWorkers);
     } else {
+      console.log('Image Workers Disabled numWorkers:', numImageWorkers);
       this.imageWorkersEnabled = false;
     }
   }
@@ -52,7 +54,7 @@ export class ImageWorkerManager {
       if (error) {
         reject(new Error(error));
       } else {
-        resolve(data);
+        resolve({ data: data as ImageBitmap });
       }
     }
   }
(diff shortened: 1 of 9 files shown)


src/core/lib/ImageWorker.ts
```diff
@@ -0,0 +1,128 @@
...
+          });
+        }
+      } catch (error) {
+        reject(error);
+      }
+    });
+  }
+}
+
+export default ImageWorkerManager;
```
frank-weindel, line 128: Can you change the default export into a direct named export?

src/core/lib/ImageWorker.ts
```diff
@@ -0,0 +1,128 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { type TextureData } from '../textures/Texture.js';
+
+type MessageCallback = [(value: any) => void, (reason: any) => void];
+
+class ImageWorkerManager {
+  isWorkerSupported = !!window.Worker;
```
frank-weindel, line 25: Use `self` instead of `window` for the case where the Core Space itself runs in a worker.

src/core/lib/ImageWorker.ts
```diff
@@ -0,0 +1,128 @@
...
+  imageWorkersEnabled = true;
+  messageManager: Record<string, MessageCallback> = {};
+  workers: Worker[] = [];
+  workerIndex = 0;
+  anchor = document.createElement('a');
```
frank-weindel, line 30: This will break Core Space in a worker (ThreadX) mode since `document` will not be available.

chiefcll: oh... anyway to fix... (thoughts) - I could do it in JS

src/core/lib/ImageWorker.ts
```diff
@@ -0,0 +1,125 @@
...
+        });
+        return { data };
+      }
+
+      self.onmessage = async (event) => {
+        const { src, premultiplyAlpha } = event.data;
+
+        try {
+          const data = await getImage(src, premultiplyAlpha);
+          self.postMessage({ src, data });
```
frank-weindel, line 78: @chiefcll Just realize that this doesn't seem to take advantage of [transfer](https://developer.mozilla.org/en-US/docs/Web/API/Worker/postMessage#transfer)?

frank-weindel:
LGTM!

frank-weindel (approved):

PR #130, opened 2024-01-22
perf: state changes causing more state changes

chiefcll:
Not invalidating the cache when x + y changes increased FPS a lot and didn't affect the app

Also I "think"
this.updateState(state); was making changes to state causing an update to be scheduled. Switch it till after to reset flag.


chiefcll:
@frank-weindel - This PR dramatically helps improvement for screens with text on it, but I'll leave it to you to research more with your work...

frank-weindel:
Replaced by #146

PR #131, opened 2024-01-23
memize shaders and timing functions

chiefcll:
Added Memize library to add a layer of caching for timing functions and Dynamic shader.

Optimized shader to return same default object which allows memize to use the object as a key efficiently. Thus most effects reuse the same props results.

### DIFF
--- a/src/core/CoreNode.ts
+++ b/src/core/CoreNode.ts
@@ -37,11 +37,7 @@ import type {
   NodeTextureLoadedPayload,
 } from '../common/CommonTypes.js';
 import { EventEmitter } from '../common/EventEmitter.js';
-import {
-  getNormalizedAlphaComponent,
-  intersectRect,
-  type Rect,
-} from './lib/utils.js';
+import { intersectRect, type Rect } from './lib/utils.js';
 import { Matrix3d } from './lib/Matrix3d.js';
 
 export interface CoreNodeProps {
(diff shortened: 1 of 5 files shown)


src/core/CoreNode.ts
```diff
@@ -351,22 +347,34 @@ export class CoreNode extends EventEmitter implements ICoreNode {
...
-        true,
-      );
-      this.premultipliedColorBr = mergeColorAlphaPremultiplied(
-        this.props.colorBr,
-        this.worldAlpha,
-        true,
-      );
 
+      // If all the colors are the same just sent them all to the same value
+      if (
```
frank-weindel, line 352: A bit surprised this would have much of an impact as these colors, unless they are changed, should only be calculated once for the node and never again during the life cycle of the node. Is it like an on-going FPS drop or just like an initial page load drop?

chiefcll: 99% of nodes are a single color - then passed to the render its split into the four props - and then the alpha color is calculated 4 times per node. 

Also these are recalculated a lot as alpha fade in out is one of the biggest animations - and I assume that gets recalculated with a new alpha value

src/core/CoreNode.ts
```diff
@@ -421,20 +429,41 @@ export class CoreNode extends EventEmitter implements ICoreNode {
...
-      'color',
-      'colorTop',
-      'colorBottom',
-      'colorLeft',
-      'colorRight',
-      'colorTl',
-      'colorTr',
-      'colorBl',
-      'colorBr',
-    ];
```
chiefcll, line 434: The biggest culprit here was probably reinitializing this array rather than storing it as a const at the top of the file.

frank-weindel (approved):
With all checks passing this LGTM

PR #138, opened 2024-01-26
Fix simple scene rendering

### LINKED ISSUE #123: Renderer doesn't render anything, unless there is at least 1 text node or a node with a src or texture
Came accross this behaviour in Blits: when the entire App contains only elements without any src or texture (but with w/h and color), nothing get's rendered on the screen.

Ways to trigger a render loop:

- Adding at least 1 element with a `src`
- Adding at least 1 element with a `texture`
- Adding a Text component (even with no actual text to render)
- Changing something in the position of an Element after the initial render

Tracked down this behaviour and it doesn't seem Blits related. It can also be reproduced in the renderer, when turning off the overlay (this removes the text node that shows on each page, which is what makes the problem apparent).

To reproduce:

the alpha test (`?test=alpha&overlay=false`) doesn't render anything. The `?test=scaling&overlay=false` test _does_ render something (as it has text and images).

My assumption is that rendering is being skipped unless a node with a `src` or `texture` flips a switch to indicate something has to be rendered on screen.

frank-weindel:
Simple scenes that do not involve textures, text or any later mutations were not rendering to the canvas. I've tracked this down to a supposed double-buffer issue involing the performance optimization that stops drawing frames when there are no updates in the render tree detected. The Renderer was actually issusing the appropriate WebGL vertices and draw calls for these simple scenes but it seems a second pass at drawing is required in order for the scene to be drawn to the canvas.

This fix simply does an extra draw/raf cycle immediately after it previously drew updates.

I looked at writing a test for this but the behavior is not exhibited when running chromium headlessly.

Fixes #123

### DIFF
--- a/src/core/platform.ts
+++ b/src/core/platform.ts
@@ -23,13 +23,16 @@ import type { Stage } from './Stage.js';
  * Platform render loop initiator
  */
 export const startLoop = (stage: Stage) => {
+  let doOneMoreRaf = true;
   const runLoop = () => {
     stage.updateAnimations();
 
-    if (!stage.hasSceneUpdates()) {
+    const hasUpdates = stage.hasSceneUpdates();
+    if (!hasUpdates && !doOneMoreRaf) {
       setTimeout(runLoop, 16.666666666666668);
       return;
     }
+    doOneMoreRaf = hasUpdates;
 
     stage.drawFrame();
     requestAnimationFrame(runLoop);


chiefcll (approved):

wouterlucas (commented):
Smells hacky, there should be no reason for this. 

We should check with the wpewebkit team why this is needed.

PR #139, opened 2024-01-26
Fix incorrect clipping rect updates when clipping parent is moved

### LINKED ISSUE #125: Incorrect child positioning with clipping property during parent translation
When the clipping property is applied to a parent node, the node added within it fails to follow the parent's position when it is moved.

```
<View width={300}    height={200} clipping>
    <View width={600} height={400} src="/image.jpg">
</View>
```

frank-weindel:
The optimization to prevent clipping updates in calculateClippingRect() when the parent clipping rect hadn't changed or the node's clipping rect was already established was erroneous and hence removed. More work will ensue later to further optimize clipping calculations which currently require a lot of allocations of objects.

Fixes #125


chiefcll (approved):

pecoram (approved):

wouterlucas (approved):
Nice one, is this changed due to the update graph behaviour?

frank-weindel:
> Nice one, is this changed due to the update graph behaviour?

Update graph behavior does guarantee that the clipping rectangle needs to be updated. Either because it's own local/global transforms were updated, the `clipping` property was toggled or because it's parent's clipping rect was updated. But I think this issue has been present since [this commit](https://github.com/lightning-js/renderer/commit/c1bcbaa65a8377cd6e10c02379dd2e4ba4f1f85b#diff-b56e593f818388b36e7083093c5bfdb07817d6b25e2aeac913c5dfb3d9cedef7R309).

PR #146, opened 2024-01-30
Perf: Optimize the setting of x/y props for SDF text

frank-weindel:
This optimization is now more optimistic about laying out text that has the potential to be on screen. With this the x/y property setters are now more optimized to not schedule an update unless the text was not renderer previously and has the potential to be rendered now.

Replaces #130


chiefcll:
Tested - nothing too scientific but seems to go from ~30fps up to ~35fps with these changes

chiefcll (approved):

Issue #151, opened 2024-02-05
Issue with .png alpha handling


wouterlucas:
PNG's can have an Alpha channel that needs to be handled with processing the texture data.

Without a fix this png:
![shines-top](https://github.com/lightning-js/renderer/assets/1534440/f7dc4db3-e7b4-4103-b1f9-c7c2c9747f3e)

will be handled as:
![Screenshot 2024-02-05 103705](https://github.com/lightning-js/renderer/assets/1534440/f679af5b-6fc7-4d77-9fcc-6bf57a4918eb)

while it should appear as:
![Screenshot 2024-02-05 103711](https://github.com/lightning-js/renderer/assets/1534440/e9314127-f60e-4b3d-89e2-a2bcc10b9433)

Setting the `premultiplyAlpha = true` on the texture data fixes this, however this will likely only need to happen when we're dealing with PNGs.

frank-weindel:
Fixed with #153

PR #153, opened 2024-02-05
fix: image load issue fix with png alpha

tuncayyildirtanuk:
Fix issue: https://github.com/lightning-js/renderer/issues/151


tuncayyildirtanuk:
I will do more tests with different scenarios.

src/core/renderers/webgl/WebGlCoreCtxTexture.ts
```diff
@@ -142,6 +142,9 @@ export class WebGlCoreCtxTexture extends CoreContextTexture {
       height = data.height;
       glw.bindTexture(this._nativeCtxTexture);
 
+      // png image loaded as expected with premultiplyAlpha set to true
+      textureData.premultiplyAlpha = true;
```
frank-weindel, line 146: For general ImageTextures the default value of `premultiplyAlpha` should be true. Curious what might be the root cause of it not being in this case.

wouterlucas: AFAIK the way Lightning 2 would do this is through a hasAlpha property that would get automatically set if the engine detected it was a PNG that would enable premultiplyAlpha. Meaning it is turned off on non-PNG images such as JPGs. Though the ImageSource and resulting TextureSource had the ability to force the alpha flag to be set through a getter/setter.

@erikhaandrikman might have more insights.

wouterlucas: So spoke to Erik and this is what he told me:
Setting the premultiplyAlph to true is indeed needed for PNG's but for JPG's its advised to be turned off. Since they do not have an alpha channel, this would benefit performance to ignore the alpha channel. By enabling this WebGL needs to process RGB times Alpha, which is more expensive to versus just the RGB channels.

I suggest we look at the extension of the image requested and toggle the alpha accordingly, just like L2 did.

frank-weindel: @wouterlucas That makes sense. We can definitely add that check at the `ImageTexture` level.

tuncayyildirtanuk:
I did some investigation. The root issue was getTextureData was not returned premultiplyAlpha param. 

So 'textureData.premultiplyAlpha' value was undefined inside 'WebGlCoreCtxTexture' for all image types.

ImageTexture changes;
1. premultiplyAlpha default value was true. I added a new 'hasAlphaChannel' method in order to define the correct premultiplyAlpha value.

2. Also, the `getTextureData` method returns the 'premultiplyAlpha' param.

I will fix typescript issue after your feedback.

frank-weindel:
Thanks for the contribution! And great 

I merged a new visual regression test that was originally missing regarding the rendering of single-channel signed distance field fonts (SSDF) #169 into main, and then merged main into this PR branch. Be sure to pull.

This PR breaks that test because we must have a way to override this new MIME-type based behavior. My suggestion is to allow the `premultiplyAlpha` property of ImageTexture to be `null | boolean`, and default it to `null`. `null` can then be interpreted to use the new MIME-type behavior to decide on how the actual premultiplyAlpha is set. If it is set explicitly to `true` or `false` that should be respected as is.

I also recommend you add the PNG image that originally caused this issue to the `alpha-blending` visual regression test (examples/tests/alpha-blending.ts). This one looks like:

alpha-blending-2

You can add it to the right of the "rocko.png" image.

I'll leave some line-by-line comments to explain this further. I'll be away on vacation for the next 3 weeks so I will likely not be able to review your final solution. But if you can follow those guidelines and make the failing test pass it should be good to go!

src/core/textures/ImageTexture.ts
```diff
@@ -74,8 +74,12 @@ export class ImageTexture extends Texture {
     this.props = ImageTexture.resolveDefaults(props);
   }
 
+  hasAlphaChannel(mimeType: string) {
+    return (mimeType.indexOf("image/png") !== -1);
+  }
+
   override async getTextureData(): Promise<TextureData> {
-    const { src, premultiplyAlpha } = this.props;
```
frank-weindel, line 78: We can't ignore the premultiplyAlpha property. If it is set explicitly to `true` or `false` it should be respected regardless of the result of `hasAlphaChannel()`.

frank-weindel:
@tuncayyildirtanuk Fixed the typescript error in: efca44453441925df1bd3c0ea739f53448dc4e82

The type was fine, but what was being put into the `resolve()` function was not the right type for that promise.

src/core/renderers/webgl/WebGlCoreCtxTexture.ts
tuncayyildirtanuk: Managed now at the ImageTexture level

src/core/textures/ImageTexture.ts
tuncayyildirtanuk: Changed as you said.

erikhaandrikman:
https://github.com/lightning-js/renderer/blob/2773aab00bd5b62bce2dbb49cb4db783a5000508/src/core/renderers/webgl/WebGlCoreCtxTexture.ts#L149

One suggestion for optimization is to update the internal format to `gl.RGB` for images without an alpha channel, this would internally use 3 bytes per pixel instead of 4.

wouterlucas:
> https://github.com/lightning-js/renderer/blob/2773aab00bd5b62bce2dbb49cb4db783a5000508/src/core/renderers/webgl/WebGlCoreCtxTexture.ts#L149
> 
> One suggestion for optimization is to update the internal format to `gl.RGB` for images without an alpha channel, this would internally use 3 bytes per pixel instead of 4.

would you suggest to do this as part of this PR or should we move this ahead and work on an optimization later?

erikhaandrikman:
> > https://github.com/lightning-js/renderer/blob/2773aab00bd5b62bce2dbb49cb4db783a5000508/src/core/renderers/webgl/WebGlCoreCtxTexture.ts#L149
> > 
> > One suggestion for optimization is to update the internal format to `gl.RGB` for images without an alpha channel, this would internally use 3 bytes per pixel instead of 4.
> 
> would you suggest to do this as part of this PR or should we move this ahead and work on an optimization later?

Lets merge this as is, and put this optimization on our backlog

wouterlucas (approved):
LGTM!

PR #158, opened 2024-02-05
Fix Canvas Text: Reported dimensions do not include descenders

### LINKED ISSUE #157: Canvas Text: Reported dimensions do not include descenders

frank-weindel:
This fix noticeably shifts canvas text upwards more when centered using the reported dimensions as it includes additional whitespace below the text now. This should be more inline with the Canvas Text implementation of Lightning 2. For this reason I'm considering this a breaking change.

Fixes #157

BREAKING CHANGE


frank-weindel:
As this does add a significant amount of padding to canvas text I'd like to dive deeper into this and see if we can perhaps reduce it.

Issue #94, opened 2023-11-30
Dom Inspector


chiefcll:
Lightning 2 Dom Inspector is a critical piece for UI testing. How can we do this with the renderer?

elsassph:
👀  and don't forget `testId`

chiefcll:
Completed - and there is a data piece that you can pass in for testId targeting

PR #167, opened 2024-02-08
Inspector: Lightning 3 node tree in the DOM for debugging

wouterlucas:
Conditional inspector that can be enabled with:
```
enableInspector: true
```

Will replicate the Lightning node tree in a DIV DOM tree, and remap Lightning 3 node properties to applicable CSS properties for easy inspection using the web inspector in your devtools.

Screenshot 2024-02-08 155139

Note: currently does not support attaching stuff like `testId` as requested in the original issue #94. This requires an extension to the `INode` interface that will need some more consideration.

Notes:
* Inspector works off of `RendererMain` so it can support both ThreadX and single threaded mode.
* Inspector wraps arround `createNode`, `createTextNode` APIs using Proxies.
* Inspector has a rudimentary understanding of animations, to reflect the DOM tree with animations being executed. However this does not work with loops and repeats at the moment.

Future considerations:
* Add a data key/value interface to the `INode`, limited to `string`, `boolean` and `number` that will be reflected back on the DOM tree in `data-*` elements.
* Listen for DevTool changes to the CSS using `MutationObserver` and reflect those onto the render node for results.

And last but not least, don't enable this in production folks. It's going to be bad mmmkay.

### DIFF
--- a/src/main-api/Inspector.ts
+++ b/src/main-api/Inspector.ts
@@ -32,9 +32,17 @@ interface StyleResponse {
   value: string;
 }
 const stylePropertyMap: {
-  [key: string]: (value: string | number | boolean) => string | StyleResponse;
+  [key: string]: (
+    value: string | number | boolean,
+  ) => string | StyleResponse | null;
 } = {
-  alpha: () => 'opacity',
+  alpha: (v) => {
+    if (v === 1) {
+      return null;
+    }
+
+    return { prop: 'opacity', value: `${v}` };
+  },
   x: (x) => {
     return { prop: 'left', value: `${x}px` };
   },
@@ -42,9 +50,17 @@ const stylePropertyMap: {
     return { prop: 'top', value: `${y}px` };
   },
   width: (w) => {
+    if (w === 0) {
+      return null;
+    }
+
     return { prop: 'width', value: `${w}px` };
   },
   height: (h) => {
+    if (h === 0) {
+      return null;
+    }
+
     return { prop: 'height', value: `${h}px` };
   },
   zIndex: () => 'zIndex',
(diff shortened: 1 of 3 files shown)


chiefcll:
Oh this is exciting. Will check it out at some point today

chiefcll:
image

Got it running the root node zIndex is - so couldnt access it under the canvas. We're also going to need a way to expose the solid node - before we had $0.element which should work again. So maybe the dom node gets returned to the framework and then we attach? Also all my nodes are not setting width and height...

src/main-api/RendererMain.ts
```diff
@@ -238,6 +239,18 @@ export interface RendererMainSettings {
...
+   * Enable inspector
+   *
+   * @remarks
+   * When enabled the renderer will spawn a inspector. The inspector will
+   * replicate the state of the Nodes created in the renderer and allow
+   * inspection of the state of the nodes.
+   *
+   * @defaultValue `false` (disabled)
+   */
+  enableInspector?: boolean;
```
chiefcll, line 253: This should be enabled via Vite build flag for dev vs production

src/main-api/Inspector.ts
```diff
@@ -0,0 +1,323 @@
...
+    // CSS mappable attribute
+    if (stylePropertyMap[property]) {
+      const mappedStyleResponse = stylePropertyMap[property]?.(value);
+
+      if (typeof mappedStyleResponse === 'string') {
+        div.style.setProperty(mappedStyleResponse, String(value));
+      } else if (typeof mappedStyleResponse === 'object') {
+        div.style.setProperty(
+          mappedStyleResponse.prop,
+          mappedStyleResponse.value,
```
chiefcll, line 264: I think you need to do width + 'px'

wouterlucas:
> image
> 
> Got it running the root node zIndex is - so couldnt access it under the canvas. We're also going to need a way to expose the solid node - before we had $0.element which should work again. So maybe the dom node gets returned to the framework and then we attach? Also all my nodes are not setting width and height...

not sure if I follow - it indeed has a zIndex underneath the canvas right now. Though on my end Chrome will still highlight the elements you mouse over and see where they are at. 
What do you mean by "expose the solid node"?

src/main-api/RendererMain.ts
frank-weindel: I wouldn't put any Vite logic in the renderer for that as I think Applications that use Vite could easily do that themselves. Also I know as a developer I wouldn't necessarily want the inspector on all the time in a development build because of its performance implications.

frank-weindel:
Nice work!

Would you be able to add a URL param option to the Example Tests to enable it? I personally wouldn't want it on by default but it'd be nice to easily be able to enable it without changing code.

src/main-api/RendererMain.ts
chiefcll: You need it to be in Vite for it to be tree shaken, otherwise its not determined till runtime and then it will always be bundled in.
https://github.com/lightning-js/renderer/blob/main/src/utils.ts#L82

wouterlucas: done for now, we can revisit later :)

michielvandergeest (approved):

PR #169, opened 2024-02-08
Add SSDF text test and snapshot

frank-weindel:
Test that was missing that should be important for #153


frank-weindel:
Just an added test. Need to expedite since I'm heading on vacation.

PR #172, opened 2024-02-13
CoreNode: Implement a cleanup routine

wouterlucas:
Implement a more rigorous cleanup logic on the CoreNode. 

Note: Does not cover ThreadX use case yet, will make separate ticket to look into ThreadX cleanup routine at a later stage.


chiefcll:
Whats the benefit of this - does removing the reference to it not clean up the node properly? Wondering performance impact...

wouterlucas:
> Whats the benefit of this - does removing the reference to it not clean up the node properly? Wondering performance impact...

Children we're certainly not cleaned up which is a bit of a pickle. Allowing dangling references to parents and thus not marking it for collection. Additionally it also forcefully clears data and marks it for GC. Allowing for early GC pickup by the engine.

One odd thing is TS won't allow me to do the variable = null; delete variable; pattern, which is known for quick GC in tests we've done before.

Note I'm sure there is more to clean when we call `destroy()` on nodes, this is a quick early pass with low fruits.

src/core/CoreNode.ts
```diff
@@ -522,6 +522,31 @@ export class CoreNode extends EventEmitter implements ICoreNode {
     this.calcZIndex = zIndex;
   }
 
+  /**
+   * Destroy the node and cleanup all resources
+   */
+  destroy(): void {
+    this.children.forEach((child) => child.destroy());
```
michielvandergeest, line 529: would a for loop be faster here?

wouterlucas: It would yes, at a certain amount of nodes it would matter. 
Will do an update, thanks for the 👀 's

wouterlucas:
Added a test, ran a test going from 10k nodes to 0 and heap size went from 40 Mb to 9 Mb:

image_2024_02_28T09_27_24_715Z

michielvandergeest (approved):
LGTM

PR #177, opened 2024-02-16
Remove PNPM warning for projects using the renderer as a dependency

michielvandergeest:
The Lightning renderer project uses PNPM as a package manager. And for anyone working on the renderer, using PNPM should obviously be required.

But for projects built on top of the renderer, there is no real reason to enforce the usage of PNPM.

Currently when installing the renderer as a dependency in an NPM based project, a warning is shown about unsupported engines. Which is not necesary and can cause confusion.

image

This change removes this warning from NPM installs using the renderer as dependency, while still enforcing PNPM when working directly on the renderer code itself.

image

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -10,3 +10,4 @@ build
 releases
 .tmp
 .env
+scripts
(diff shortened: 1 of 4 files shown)


wouterlucas (approved):
Nice 👍

erikhaandrikman (approved):

PR #178, opened 2024-02-20
Feat: Allow custom data key/value storage on Nodes.

wouterlucas:
This allows the consumer of the Node's or TextNode's to inject small data properties that will be returned and/or added in the inspector `node-*` attributes.

For example:
```
  const dataNode = renderer.createNode({
    width: 505,
    height: 101,
    x: 200,
    y: 200,
    src: logo,
    parent: bg,
    data: {
      id :'dataNode',
      number: 1,
      boolean: true
    }
  });
```

In this case data: {} is an santized Map of key/values that will be stored on the main node.

*Note*: Custom data will not be passed to the CoreNode or ThreadX node. Thus data remains on the main thread and does not get forwarded to the Render thread in case of the ThreadX driver. This to avoid clashes/overhead on the ThreadX interface.

For performance reasons only strings, booleans and numbers are allowed. Passing in a nested object will not be permitted to avoid burdening Lightning nodes with excessive data.

Relates to https://github.com/lightning-js/renderer/pull/167 for `data-*` in the Inspector output.

### DIFF
--- a/src/render-drivers/utils.ts
+++ b/src/render-drivers/utils.ts
@@ -58,21 +58,34 @@ export async function loadCoreExtension(
 }
 
 export function santizeCustomDataMap(d: CustomDataMap): CustomDataMap {
-  for (const key in d) {
-    const value = d[key];
+  const validTypes = { boolean: true, string: true, number: true };
 
-    if (typeof value === 'string' && value.length > 2048) {
+  const keys = Object.keys(d);
+  for (let i = 0; i < keys.length; i++) {
+    const key = keys[i];
+    if (!key) {
+      continue;
+    }
+
+    const value = d[key];
+    const valueType = typeof value;
+
+    // Typescript doesn't understand the above const valueType ¯\_(ツ)_/¯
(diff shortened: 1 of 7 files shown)


chiefcll (approved):

src/render-drivers/utils.ts
```diff
@@ -55,3 +56,29 @@ export async function loadCoreExtension(
...
+      console.warn(
+        `Custom Data value for ${key} is too long, it will be truncated to 2048 characters`,
+      );
+      d[key] = value.substring(0, 2048);
+    }
+
+    if (
+      typeof value !== 'boolean' &&
+      typeof value !== 'string' &&
+      typeof value !== 'number'
```
michielvandergeest, line 74: Wouldn't it make sense to do the `typeof` check once, store that in a const like you do with `const value = d[key];`?

Also, I would probably flip the 2 if statements and do an early return.

Both micro optimizations perhaps, but since this potentially happens for many new nodes it may pay off ..

PR #184, opened 2024-02-24
fix: #28 Set height + width to calculated

chiefcll:
When loading text set the width and height of the text to the calculated width and height depending on contain modes. Thus when users ask for the width and height they get the actual values rather than the props that they originally set which aren't relevant.


chiefcll:
PS - Currently in solid we get the width + height on text load and then set it back to the renderer which causes a double updateState to occur on the text.

wouterlucas:
This seems to break two visual regression tests:
```
Running clipping-2... FAILED!
Running text-contain-3... FAILED!
```

can you check what's going on?

chiefcll:
Was waiting for @frank-weindel to chime in...

frank-weindel:
Replacing with #201

PR #185, opened 2024-02-24
feat: add font weight fallbacks

chiefcll:
Update the font resolver to return a font for different fontWeights. If a fontWeight is not available it will get a font based on the CSS rules: https://www.w3.org/TR/2018/REC-css-fonts-3-20180920/#font-matching-algorithm

Also memize the function for performance

### DIFF
--- a/src/core/text-rendering/TrFontManager.ts
+++ b/src/core/text-rendering/TrFontManager.ts
@@ -17,10 +17,105 @@
  * limitations under the License.
  */
 
-import { SdfTrFontFace } from './font-face-types/SdfTrFontFace/SdfTrFontFace.js';
 import type { TrFontFace } from './font-face-types/TrFontFace.js';
-import { WebTrFontFace } from './font-face-types/WebTrFontFace.js';
 import type { TextRendererMap, TrFontProps } from './renderers/TextRenderer.js';
+import memize from 'memize';
+
+const weightConversions: { [key: string]: number } = {
+  normal: 400,
+  bold: 700,
+  bolder: 900,
+  lighter: 100,
+};
+
+const fontWeightToNumber = (weight: string | number): number => {
+  if (typeof weight === 'number') {
+    return weight;
+  }
+
+  return weightConversions[weight] || 400;
+};
+
+function rawResolveFontToUse(
+  familyMapsByPriority: FontFamilyMap[],
+  family: string,
+  weightIn: string | number,
+  style: string,
+  stretch: string,
+): TrFontFace | undefined {
+  const weight = fontWeightToNumber(weightIn);
+  return familyMapsByPriority.reduce<TrFontFace | undefined>(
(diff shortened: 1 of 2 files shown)


src/core/text-rendering/TrFontManager.ts
```diff
@@ -17,10 +17,105 @@
...
+
+function rawResolveFontToUse(
+  familyMapsByPriority: FontFamilyMap[],
+  family: string,
+  weightIn: string | number,
+  style: string,
+  stretch: string,
+): TrFontFace | undefined {
+  const weight = fontWeightToNumber(weightIn);
+  return familyMapsByPriority.reduce<TrFontFace | undefined>(
```
wouterlucas, line 47: consider using an for loop over reduce, reduce is a lot slower:
https://leanylabs.com/blog/js-forEach-map-reduce-vs-for-for_of/

src/core/text-rendering/TrFontManager.ts
```diff
@@ -17,10 +17,105 @@
...
+
+      const fontFaces = fontFamiles[family];
+      if (!fontFaces) {
+        return undefined;
+      }
+
+      if (fontFaces.size === 1) {
+        // No Exact match found, find nearest weight match
+        const msg = `TrFontManager: Only one font face found for family: '${family}' - will be used for all weights and styles`;
+        console.warn(msg);
```
wouterlucas, line 61: why stick it in a const first if youre just going to print it? wouldn't a straight literal to console.war() work too?
Saves another const :)

src/core/text-rendering/TrFontManager.ts
```diff
@@ -17,10 +17,105 @@
...
+          return bolder;
+        }
+      }
+
+      return undefined;
+    },
+    undefined,
+  );
+}
+const resolveFontToUse = memize(rawResolveFontToUse);
```
wouterlucas, line 118: does this get called that often with different params that caching is required here? I though this would only get called on init? 
Is the extra memory consumption worth the performance?

I'd argue if the function is relatively short & simple the JIT compiler will optimize this beyond any caching you can do manually at the cost of something that won't be GC'ed.

wouterlucas (commented):
Do all the tests cover this? or do we need to add/update a visual regression test for this?

src/core/text-rendering/TrFontManager.ts
chiefcll: This is what was there... But I think it could be a little simpler - reduce always confuses me 😢

src/core/text-rendering/TrFontManager.ts
chiefcll: This gets called for every textNode that is created as everyone requires a fontFile to be determined for it. Most apps use ~3 fonts - So I think memize this is a good thing. Will run some tests at some point on how well the JIT does with it

PR #186, opened 2024-02-28
Render to Texture

erikhaandrikman:
This feature introduces the capability to render the scene onto a texture, facilitating post-processing effects and potentially allowing for the caching of static UI elements.

### Tasks

- [x] Nested render textures follow a bottom-up render order
- [x] Automatic clipping support
- [x] Rendering framebuffer smaller than the viewport maintains correct aspect ratio

### Done

- [x] Rendering the node and its children onto a texture
- [x] Updating the render texture to reflect changes in child nodes
- [x] Prevent buffering of render texture child vertices


wouterlucas:
lets go

frank-weindel:
Super cool!!! Love the reflection demo.

src/core/renderers/webgl/WebGlCoreRenderer.ts
```diff
@@ -450,13 +476,13 @@ export class WebGlCoreRenderer extends CoreRenderer {
...
   private addTexture(
-    texture: WebGlCoreCtxTexture,
+    texture: WebGlCoreCtxTexture | WebGLTexture,
     bufferIdx: number,
     recursive?: boolean,
   ): number {
     const { curRenderOp } = this;
     assertTruthy(curRenderOp);
-    const textureIdx = curRenderOp.addTexture(texture);
+    const textureIdx = curRenderOp.addTexture(texture as WebGlCoreCtxTexture);
```
frank-weindel, line 485: This assertion and the added `WebGLTexture` type to the `texture` parameter is causing a `RenderTexture` to be passed through to a render op in a unsafe way.

src/core/textures/RenderTexture.ts
```diff
@@ -0,0 +1,138 @@
...
+  width?: number;
+
+  /**
+   * WebGL Texture height
+   * @default 256
+   */
+  height?: number;
+}
+
+export class RenderTexture extends Texture {
```
frank-weindel, line 41: This class seems to combine aspects of `WebGlCoreTexture` and `Texture` with added WebGL code for render-to-texture. `Texture` is intended to be non-platform specific with no WebGL specific code. WebGL specific code would better live in an extension to `WebGlCoreTexture` instead. Having this class extend `WebGlCoreTexture` might also automatically resolve the typing issue I reported in my [other comment](https://github.com/lightning-js/renderer/pull/186/files#r1513321615).

src/main-api/INode.ts
```diff
@@ -405,6 +405,21 @@ export interface INodeWritableProps {
...
+   * TBD
+   *
+   * @default false
+   */
+  rtt: boolean;
+
+  /**
+   * Whether the Node has a Render Texture as a parent
+   */
+  parentHasRenderTexture: boolean;
```
frank-weindel, line 422: From what I can gather this isn't a property that is ever set by a user. If that's the case, it doesn't belong in this interface. I think you can actually keep this purely within the Core space (i.e CoreNode).

src/render-drivers/main/MainOnlyNode.ts
```diff
@@ -117,10 +119,11 @@ export class MainOnlyNode extends EventEmitter implements INode {
     this.coreNode.on('inViewport', this.onInViewport);
 
     // Assign properties to this object
-    this.parent = props.parent as MainOnlyNode;
+    this.parent = props.parent as unknown as MainOnlyNode;
```
frank-weindel, line 122: Is this change necessary? I tried switching it back on my machine and didn't get any errors.

src/core/CoreNode.ts
```diff
@@ -1235,9 +1294,73 @@ export class CoreNode extends EventEmitter implements ICoreNode {
...
+   * Returns the framebuffer dimensions of the node.
+   * If the node has a render texture, the dimensions are the same as the node's dimensions.
+   * If the node does not have a render texture, the dimensions are inherited from the parent.
+   * If the node parent has a render texture and the node is a render texture, the nodes dimensions are used.
+   */
+  get framebufferDimensions(): Dimensions | undefined {
+    if (this.parentHasRenderTexture && !this.rtt) {
+      return this.parent?.framebufferDimensions;
+    }
+    return { width: this.width, height: this.height };
```
frank-weindel, line 1340: If a user updates the width/height properties of an RTT Node this may cause the frame buffer dimensions to be reported differently than the size of the texture that is initially allocated.

src/core/CoreNode.ts
```diff
@@ -1235,9 +1294,73 @@ export class CoreNode extends EventEmitter implements ICoreNode {
...
+    return this.props.parentHasRenderTexture;
+  }
+
+  /**
+   * Returns the framebuffer dimensions of the node.
+   * If the node has a render texture, the dimensions are the same as the node's dimensions.
+   * If the node does not have a render texture, the dimensions are inherited from the parent.
+   * If the node parent has a render texture and the node is a render texture, the nodes dimensions are used.
+   */
+  get framebufferDimensions(): Dimensions | undefined {
```
frank-weindel, line 1336: Technically `undefined` will never happen since `this.parent?` will always resolve. You could indicate this with `this.parent!` and remove the `undefined`.

src/core/CoreNode.ts
```diff
@@ -1235,9 +1294,73 @@ export class CoreNode extends EventEmitter implements ICoreNode {
...
+    // Store RTT nodes in a separate list
+    this.stage.renderer?.renderToTexture(this);
+  }
+
+  set parentHasRenderTexture(value: boolean | undefined) {
+    this.props.parentHasRenderTexture = !!value;
+  }
+
+  get parentHasRenderTexture(): boolean {
+    return this.props.parentHasRenderTexture;
```
frank-weindel, line 1327: As mentioned in another comment, this property could just live in the CoreNode and simply reach up the parent tree to find out if a parent has RTT on. If that reaching up is inefficient it could be calculated during an update() stage.

src/core/CoreNode.ts
```diff
@@ -1235,9 +1294,73 @@ export class CoreNode extends EventEmitter implements ICoreNode {
...
+
+  set rtt(value: boolean) {
+    if (!value) {
+      return;
+    }
+    this.props.rtt = true;
+    this.hasRTTupdates = true;
+
+    // Store RTT nodes in a separate list
+    this.stage.renderer?.renderToTexture(this);
```
frank-weindel, line 1319: If `rtt` is set to `false` does it remain in this structure?

examples/tests/rtt.ts
```diff
@@ -0,0 +1,89 @@
...
+import test from './alpha-blending.js';
+interface AnimationExampleSettings {
+  duration: number;
+  easing: string;
+  delay: number;
+  loop: boolean;
+  stopMethod: 'reverse' | 'reset' | false;
+}
+
+const animationSettings: Partial<AnimationExampleSettings> = {
```
frank-weindel, line 11: Could you turn one of these example tests into a visual regression test? Animations wouldn't really be necessary to test in them obviously.

src/core/CoreNode.ts
```diff
@@ -321,13 +331,22 @@ export class CoreNode extends EventEmitter implements ICoreNode {
...
+
     this.emit('loaded', {
       type: 'texture',
       dimensions,
     } satisfies NodeTextureLoadedPayload);
+
+    queueMicrotask(() => {
+      // Texture was loaded. In case the RAF loop has already stopped, we request
+      // a render to ensure the texture is rendered.
+      this.stage.requestRender();
```
frank-weindel, line 348: Any reason this needs to be in a microtask?  All it does is set a boolean flag.

src/core/renderers/webgl/WebGlCoreRenderer.ts
```diff
@@ -486,6 +494,50 @@ export class WebGlCoreRenderer extends CoreRenderer {
     return textureIdx;
   }
 
+  /**
+   * Test if the current Render operation can be reused for the specified parameters.
+   * @param params
+   * @returns
+   */
+  reuseRenderOp(params: QuadOptions) {
```
frank-weindel, line 502: Nice. Much cleaner as a separate function.

src/core/CoreNode.ts
erikhaandrikman: Dimension updates will now also reflect to the render texture

src/core/CoreNode.ts
erikhaandrikman: Good question!  It prompted me to revisit this. I've added support to dynamically toggle render to texture on a Node. This opens up the door for rendering an entire app to a texture, allowing for the application of post-processing effects, before disabling render to texture.

src/core/CoreNode.ts
erikhaandrikman: Moved the logic for tracking whether the parent has a render texture to the CoreNode. This is now being calculated during node `update()`

examples/tests/rtt.ts
erikhaandrikman: done!

frank-weindel (approved):

frank-weindel:
LGTM! Awesome work.

Issue #180, opened 2024-02-22
Inspector Updates


chiefcll:
1. zIndex on root should be higher than canvas tag so it's over it to make it easier to inspect elements
  - TextNodes should have the text but text-indent it off screen
  - Background images - do they need to be on there? Maybe just a data-src rather than actually loading them
3. support testId or some way to label the nodes to make it easier to get access to them
4. Node needs to contain a reference to the div that its associated with - Frameworks need this to add their own props to allow users access to their "framework nodes"

@wouterlucas - 4 is the most important - I can't use this until there is some way to get the solid node with the dom tree.

wouterlucas:
@chiefcll thanks for the feedback!

1. Yeah I initially had a lower zIndex because of the text (which doesnt understand SDF) and images obstructing it. I guess doing it like that works better and allows you to right click it. Though in today's case the current position and such already show up highlighted in chrome dev tools. I'll try to align it.

3. Already there, just need this to be merged in first: https://github.com/lightning-js/renderer/pull/178 - that way anything in the the `data : {}` object will be exposed as `data-*` attribute on the DOM element.

4. It would create a circular reference though, node pointing to a div pointing to a node. I wont be able to expose solid nodes as the Renderer has no access to that. Wouldn't the `data` object be sufficient? Or could you provide a use case on how this would be used?

chiefcll:
1. The big feature is being able to right click on a div and it match up with the UI - so zIndex is important and if alpha:0 then display: none the div

3. If the node that you return has the id that you created with it I could getElementById to attach the solid node to it so that would work for me (just a bit extra cost, but I assume you wouldn't be able to pass dom reference via threadx)

chiefcll:
Working now in solid 0.13.2

PR #187, opened 2024-02-28
Inspector: Put divs on top, attach div to node and handle images differently

wouterlucas:
Feedback from https://github.com/lightning-js/renderer/issues/180

### DIFF
--- a/src/main-api/Inspector.ts
+++ b/src/main-api/Inspector.ts
@@ -111,13 +111,6 @@ const stylePropertyMap: {
 
     return { prop: 'transform', value: `scaleY(${v})` };
   },
-  src: (v) => {
-    if (!v) {
-      return null;
-    }
-
-    return { prop: 'background-image', value: `url(${v})` };
-  },
   color: (v) => {
     if (v === 0) {
       return null;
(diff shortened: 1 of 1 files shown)


src/main-api/Inspector.ts
```diff
@@ -221,7 +214,7 @@ export class Inspector {
     this.root.style.transformOrigin = '0 0 0';
     this.root.style.transform = `scale(${this.scaleX}, ${this.scaleY})`;
     this.root.style.overflow = 'hidden';
-    this.root.style.zIndex = '-65534';
+    this.root.style.zIndex = '65534';
```
michielvandergeest, line 217: Do we really need to set an explicit z-index (and the highest possible z-index)? Via natural z index ordering the inspector div already overlays the canvas simply for being inserted after the canvas.

chiefcll: Concur.

wouterlucas: It’s what the L2 inspector did too:
https://github.com/rdkcentral/Lightning/blob/master/devtools/lightning-inspect.js#L220

Likely to avoid anything else obstructing it, like video players or other DOM elements. 

Note this is only for the root element, rest is sorted per node zIndex. 

Honestly I don’t think this hurts nor matters, having the inspector at absolute top if ‘right clicking’ the elements is such an important use case by @chiefcll ’s suggestion, then this will make sure it always works.

michielvandergeest (approved):
LGTM

chiefcll (approved):

PR #188, opened 2024-03-02
fix: warning for font weight

chiefcll:
should've left it with the reduce 😭

### DIFF
--- a/src/core/text-rendering/TrFontManager.ts
+++ b/src/core/text-rendering/TrFontManager.ts
@@ -106,6 +106,15 @@ function rawResolveFontToUse(
       }
       weight += 100;
     }
+
+    // finally check lower again
+    weight = 500;
+    while (weight > 0) {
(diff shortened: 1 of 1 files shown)


src/core/text-rendering/TrFontManager.ts
```diff
@@ -85,35 +78,35 @@ function rawResolveFontToUse(
...
         weightMap.get(900);
       if (bolder) {
-        result = bolder;
+        return bolder;
       }
     }
   }
 
-  return result;
+  return undefined;
```
wouterlucas, line 109: i dont think you need to return an explicit undefined here, just doing return without anything would do the same.

wouterlucas (commented):
what warning does this fix?

chiefcll:
Every font that is loaded gets warning - it should've returned

frank-weindel (approved):

PR #191, opened 2024-03-07
Feat/render bounds

jfboeve:
This pull request contains an implementation that excludes nodes that are located outside of the clippingrect/viewport from being added as quads. You can also add boundsMargin in order to extend the area in which nodes will still be loaded as quads

### DIFF
--- a/src/core/CoreNode.ts
+++ b/src/core/CoreNode.ts
@@ -55,6 +55,12 @@ export enum CoreNodeRenderState {
   InViewport = 8,
 }
 
+const CoreNodeRenderStateMap: Map<CoreNodeRenderState, string> = new Map();
+CoreNodeRenderStateMap.set(CoreNodeRenderState.Init, 'init');
+CoreNodeRenderStateMap.set(CoreNodeRenderState.OutOfBounds, 'outOfBounds');
+CoreNodeRenderStateMap.set(CoreNodeRenderState.InBounds, 'inBounds');
+CoreNodeRenderStateMap.set(CoreNodeRenderState.InViewport, 'inViewport');
+
 export interface CoreNodeProps {
   id: number;
   // External facing properties whose defaults are determined by
@@ -343,13 +349,13 @@ export class CoreNode extends EventEmitter implements ICoreNode {
         this.globalTransform.multiply(this.localTransform);
       }
       this.calculateRenderCoords();
+      this.updateRenderState();
       this.setUpdateType(UpdateType.Clipping | UpdateType.Children);
       childUpdateType |= UpdateType.Global;
     }
 
     if (this.updateType & UpdateType.Clipping) {
       this.calculateClippingRect(parentClippingRect);
-      this.updateRenderState();
       this.setUpdateType(UpdateType.Children);
       childUpdateType |= UpdateType.Clipping;
     }
@@ -527,9 +533,14 @@ export class CoreNode extends EventEmitter implements ICoreNode {
       const previous = this.renderState;
       this.renderState = renderState;
       if (previous === CoreNodeRenderState.InViewport) {
-        this.emit('OutOfViewport');
+        this.emit('outOfViewPort', {
+          previous,
+          current: renderState,
+        });
       }
-      this.emit(CoreNodeRenderState[renderState], {
+      const event = CoreNodeRenderStateMap.get(renderState);
+      assertTruthy(event);
+      this.emit(event, {
         previous,
         current: renderState,
       });
--- a/src/core/CoreNode.ts
+++ b/src/core/CoreNode.ts
@@ -408,6 +408,7 @@ export class CoreNode extends EventEmitter implements ICoreNode {
           true,
         );
       }
+      this.updateIsRenderable();
       this.setUpdateType(UpdateType.Children);
       childUpdateType |= UpdateType.PremultipliedColors;
     }
@@ -497,7 +498,6 @@ export class CoreNode extends EventEmitter implements ICoreNode {
     if (this.props.colorBr !== 0) {
       return true;
     }
-
     return false;
   }
 
@@ -565,7 +565,7 @@ export class CoreNode extends EventEmitter implements ICoreNode {
   // It returns true if any of the specified properties are truthy or if any color property is not 0, otherwise it returns false.
   updateIsRenderable() {
     if (!this.checkRenderProps()) {
-      this.isRenderable = false;
+      return (this.isRenderable = false);
     }
     this.isRenderable = this.renderState > CoreNodeRenderState.OutOfBounds;
   }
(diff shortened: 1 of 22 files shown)


wouterlucas:
Squashed a bit of the commit log and fixed the test. @frank-weindel I'd imagine this would need an `INode` extension too to support `onOutOfViewport` / `onInViewport` too for my editor to stop complaining about it?

image

frank-weindel:
> Squashed a bit of the commit log and fixed the test. @frank-weindel I'd imagine this would need an `INode` extension too to support `onOutOfViewport` / `onInViewport` too for my editor to stop complaining about it?
> 
> image

Yep. But furthermore we should use the event emitter for this. This is exactly the kind of thing it is designed for.

jfboeve:
So the reason I did this is because right now we are emitting emits... which I'm not sure if its the way to go ? With this solution you can just hook directly to the node and only add a listener when a user actually need one.

src/render-drivers/main/MainOnlyNode.ts
```diff
@@ -116,6 +116,22 @@ export class MainOnlyNode extends EventEmitter implements INode {
     this._data = props.data;
   }
 
+  onOutOfBounds(fn: () => void) {
+    this.coreNode.on('OutOfBounds', fn);
```
frank-weindel, line 120: All events names should begin with a lower case letter to remain consistent with existing events: `outOfBounds`, `inBounds`, `outOfViewport`, `inViewport`

src/core/text-rendering/renderers/TextRenderer.ts
```diff
@@ -20,6 +20,7 @@
 import type { EventEmitter } from '../../../common/EventEmitter.js';
 import type { Stage } from '../../Stage.js';
 import type { Matrix3d } from '../../lib/Matrix3d.js';
+import type { RenderCoords } from '../../lib/RenderCoords.js';
```
frank-weindel, line 23: Currently unused. Do you have a plan to use it?

src/core/text-rendering/renderers/SdfTextRenderer/SdfTextRenderer.ts
```diff
@@ -56,6 +56,7 @@ import type {
 import type { WebGlCoreCtxTexture } from '../../../renderers/webgl/WebGlCoreCtxTexture.js';
 import { EventEmitter } from '../../../../common/EventEmitter.js';
 import type { Matrix3d } from '../../../lib/Matrix3d.js';
+import type { RenderCoords } from '../../../lib/RenderCoords.js';
```
frank-weindel, line 59: Currently unused. Do you have a plan to use it?

frank-weindel (changes requested):
See comments

src/render-drivers/main/MainOnlyNode.ts
jfboeve: changed the renderstate events to fall in line with the way Texture events are emitted.

src/core/CoreNode.ts
```diff
@@ -324,14 +342,14 @@ export class CoreNode extends EventEmitter implements ICoreNode {
...
-
+      this.calculateRenderCoords();
       this.setUpdateType(UpdateType.Clipping | UpdateType.Children);
       childUpdateType |= UpdateType.Global;
     }
 
     if (this.updateType & UpdateType.Clipping) {
       this.calculateClippingRect(parentClippingRect);
-      this.checkIsRenderable();
+      this.updateRenderState();
```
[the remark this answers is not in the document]
jfboeve: moved this call, to after the render coords are made.

src/core/text-rendering/renderers/TextRenderer.ts
jfboeve: cleaned up

src/core/text-rendering/renderers/SdfTextRenderer/SdfTextRenderer.ts
jfboeve: cleaned up

src/core/CoreNode.ts
```diff
@@ -415,64 +432,190 @@ export class CoreNode extends EventEmitter implements ICoreNode {
...
+    const { width, height, globalTransform: transform } = this;
+    assertTruthy(transform);
+    const { tx, ty, ta, tb, tc, td } = transform;
+    if (tb === 0 && tc === 0) {
+      const minX = tx;
+      const maxX = tx + width * ta;
+
+      const minY = ty;
+      const maxY = ty + height * td;
+      this.renderCoords = new RenderCoords([
```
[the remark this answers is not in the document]
jfboeve: Actually for the CoreNode / CoreRenderer I want to keep the `RenderCoords` for a different optimalisation. where we pass the `RenderCoords` to the CoreRenderer `addQuad` function. This would mean that we won't have to calculate the coords again which is happening now every time a node is added as quad.

However for this to properly work there are some changes needed in the TextCanvasRenderer which I plan to do in a different PR as it is unrelated to this feature.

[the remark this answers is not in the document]
frank-weindel: Gotcha. I think I see where you are going with it and it makes sense. I'd then just suggest some optimizations to maybe cut down on array allocations and maybe reusing a single RenderCoords instance, similar to how its done for matrices and some rect/bounds instances.

frank-weindel:
> So the reason I did this is because right now we are emitting emits... which I'm not sure if its the way to go ? With this solution you can just hook directly to the node and only add a listener when a user actually need one.

It is the way it is now. If we want to optimize that we need to look at it holistically. I personally rather not have a system that provides an EventEmitter AND specific event hook methods. It should be one or the other. EventEmitter provides lots of tools. It is possible we could improve the EventEmitter by not hooking up events that aren't being ultimately listened for.

src/core/CoreNode.ts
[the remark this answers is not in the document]
jfboeve: I've updated the RenderCoords and some Bound instances!

frank-weindel:
This looks pretty good to me now! I just wish we had an automated way of testing it. It might be possible in a visual regression test using text to verify that the appropriate events fire when moving a node in and out of the render bounds/viewport.

wouterlucas (approved):
LGTM!

wouterlucas:
> This looks pretty good to me now! I just wish we had an automated way of testing it. It might be possible in a visual regression test using text to verify that the appropriate events fire when moving a node in and out of the render bounds/viewport. Curious if you have any thoughts on it.

could probably do a vitetest unit test around it without actually rendering something? Stub the addQuads part and run some dummy coordinates/positions? Thought not a quick thing to set up.

frank-weindel:
> could probably do a vitetest unit test around it without actually rendering something? Stub the addQuads part and run some dummy coordinates/positions? Thought not a quick thing to set up.

Right, not quick to set up. I figured with a VRT test you're testing the whole pipeline.

frank-weindel (approved):

Issue #192, opened 2024-03-11
createImageBitmap not supported on older browsers


michielvandergeest:
The Lightning renderer depends on `createImageBitmap` webAPI for loading images, but this webAPI isn't available in browsers older than Chrome 59.

Lightning 2 used to have a fallback for when `createImageBitmap` wasn't available, so it probably make sense to add this feature in Lightning 3 renderer as well.

PR #194, opened 2024-03-11
README: Add browser targets section

frank-weindel:
Aiming to provide solid clarity on what Lightning 3's goal of browser support is.

Added the Chrome v38 requirement after discussing with @wouterlucas. If we know for sure any other browser versions (and dates) let's discuss here so we can add them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,14 @@ pnpm start
 pnpm start:prod
 ```
 
+## Browser Targets
+
+The Lightning 3 Renderer's goal is to work with the following browser versions and above:
+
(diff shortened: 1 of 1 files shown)


wouterlucas:
We'll probably need to add some text on how to achieve running on older browser engines, thinking of:
* Set target in the vite config
* Disable image workers if Chrome <= 50 (due to createImageBitmap)
* And needs this to be resolved: https://github.com/lightning-js/renderer/issues/192

wouterlucas (approved):

PR #195, opened 2024-03-12
fix #171

### LINKED ISSUE #171: Text align in canvas renderer doesn't work for single line text
When using the canvas text renderer the property `textAlign` doesn't take effect for texts that fit on a single line. Only when there is more text than the specified `width` the text is centered or aligned to the right.

Steps to reproduce:

- use the existing `examples/tests/text.ts`
- change the text on L88 to something shorter (`Hello world`)
- add `textAlign: 'center'` to `initialProps` object on L90

Observe that the text is _not_ centered (using canvas. sdf mode _does_ center the text).

- Add more text on L88 (enough to overflow the width)

Observe that the _second line of text_ now is correctly centered (the first one is still center left, though).

See screenshot below:

image

image

image

image

Drulokia:
fix #171 Text align in canvas renderer doesn't work for single line text


frank-weindel:
Replaced by #265. I included you as a co-author though because your it's only a slight modification of your solution (+ a test)

PR #199, opened 2024-03-13
Test update of snapshots from M3 MacBook

frank-weindel:


wouterlucas (approved):
LGTM :)

frank-weindel:
@wouterlucas Nah this was me trying to figure out what was going on. The fix is not straight forward. Strangely the `destroy` tests are failing here because the entire background is 1 unit of red different depending on if its captures on Intel Mac vs M3...

PR #200, opened 2024-03-14
fix #198 - delete MainOnlyNodes and children

### LINKED ISSUE #198: Memory Leak
I'm researching a memory leak with App. At this point I assume it's in the Renderer based on this:
When I load the TV Grid I have an equal number of ElementNodes (my solid nodes) as Renderer Nodes (Core + Text)
Screenshot 2024-03-13 at 11 46 22 AM

As I navigate a TV grid, creating and removing dozens of nodes per row I end up with roughly the same element nodes as I started with, but the Renderer Nodes are in the thousands. I am calling node.destroy() on the parent node... 

Screenshot 2024-03-13 at 11 45 27 AM

chiefcll:


chiefcll:
Change works well for me:

image

wouterlucas (changes requested):
I did have this in staged changes, however during my tests I couldn't find a noticeable difference in memory cleanup over running the delete off the CoreNode so decided not to do anything with it. If this does help in a _real world_ case that all the merrier.

However, you're now cleaning up `MainOnlyNodes` and recursively cleanup `MainOnlyNode` children. The `CoreNode` will still have children as we're only unsetting the parent on node `destroy()`. This means that, although disconnected from the parent, each `CoreNode` will still have an array of children with nodes that could potentially leak. 

In my previous rationale, since `CoreNodes` have the bigger impact on memory, I sticked to that.

So to make this fully complete, either _also_ cleanup the CoreNode children or update the `set parent` function in `CoreNode` to include handling a `null` case where we remove the node from its "oldParent" childrenlist so its dereferenced. 

That and this would need a ThreadX Node solution too, or a ticket to track that so we won't forget about it in the future.

chiefcll:
Left the removal in CoreNode as well. @frank-weindel can you update the VRT for me please 🙏

frank-weindel:
> However, you're now cleaning up `MainOnlyNodes` and recursively cleanup `MainOnlyNode` children. The `CoreNode` will still have children as we're only unsetting the parent on node `destroy()`. This means that, although disconnected from the parent, each `CoreNode` will still have an array of children with nodes that could potentially leak.

Each `MainOnlyNode` that you set its parent to `null` on also causes its corresponding `CoreNode` parent to be set to `null`. And both have handlers in this case to remove themselves from their old parent's `children` array. So I'd fully expect just doing the tree traversal on the `MainOnlyNode` (and `ThreadXMainNode`) to clean up everything.

Am I missing something perhaps?

wouterlucas:
> Each `MainOnlyNode` that you set its parent to `null` on also causes its corresponding `CoreNode` parent to be set to `null`. And both have handlers in this case to remove themselves from their old parent's `children` array. So I'd fully expect just doing the tree traversal on the `MainOnlyNode` (and `ThreadXMainNode`) to clean up everything.
> 
> Am I missing something perhaps?

No you're absolutely right, I've overlooked the cleanup on the `set parent`. We might just move the:
`this.parent.setUpdateType(UpdateType.Children);` down to where we update the parents children for other cases when we `null` the parent of a node.

Also might be good to check for a `null` already being set in that section. I believe we ran into that when Blits tried to already `null` the parent prior to calling `destroy` on the node.

wouterlucas (approved):

frank-weindel (approved):

Issue #183, opened 2024-02-23
Sdf  maxLines not working with contain: "both"


pecoram:
Using SDF maxLines works only with contain: "width" and not with contain: "both"

```
width: 870,
contain: "both",
maxLines: 2,
```
![image](https://github.com/lightning-js/renderer/assets/33352553/ffa456d4-022d-43c5-8604-a72342a79be2)

```
width: 870,
contain: "width",
maxLines: 2,
```
![image](https://github.com/lightning-js/renderer/assets/33352553/da738fb2-66db-48f1-b9cf-18ac97f7f292)

chiefcll:
I saw this issue as well - not sure if it's valid though. Contain: both implies that you want it within a width and height, but you don't have a set height. Maxlines says draw 2 lines worth of text - We know height = lineHeight or fontSize * 2. 

What are you expecting in this case? I think maxlines should always 'win' and two lines should be rendered

pecoram:
So the issue results in the fact that Canvas and SDF have two different behaviors

Issue #170, opened 2024-02-09
Text Rendering: Reported Text Height Change Proposal


frank-weindel:
The following proposal aims to make the Reported Text Height, which is provided during the `loaded` event for Text Nodes, consistent between SDF and Canvas text renderers, closer to how HTML/CSS works, and solve some issues that have been reported around the current implementation:
- #165
- #157 

The phases are divided as such to make it easier to grok the issues and their solutions. We could very well implement and release all of phases at once.

## Phase 1: Make Reported Text Height dimension of text based purely on Line Height and number of lines rendered (plus offsetY)
- Reasoning
  - This is how browsers reports text dimensions for inline-block HTML div tags containing only text.
  - This will result in consistent Height reporting regardless of which characters are used on the last line of text.
- Prior to this phase, Reported Text Height was based on:
	- SDF: The actual Y axis extents of the rendered characters.
		- The problem with this is that the presence/absence of descenders (parts of glyphs that hang below the baseline such as in: jqpg) on the last line of text will affect the Reported Text Height. This is generally NOT desired when the Reported Text Height is used for vertically centering text or determining the Y position of the next node in a vertical layout.
	- Canvas (Currently):  Line Height times number of lines
		- **Check if offsetY needs be added, otherwise no change needed
	- Canvas (Renderer PR #158): This behavior, copied from Lightning 2, uses the following formula to calculate the Reported Text Height
		- This formula is mostly based on lineHeight+numLines+offsetY, except for the last line it attempts to ensure the text will fit by adding additional, somewhat arbitrarily defined, pixels.
		```ts
		const baselineOffset = textBaseline !== 'bottom' ? 0.5 * fontSize : 0;
		return (
		  lineHeight * (numLines - 1) +
		  baselineOffset +
		  Math.max(lineHeight, fontSize) +
		  (offsetY || 0)
		);
		```
- If the Line Height is too small there's a chance for characters to bleed out of the reported text dimensions.
	- For Phase 1 the Default Line Height will remain the same as the Font Size. This is NOT always enough to guarantee rendered text will fit within the Reported Text Height. Phase 2 ensures that a Minimum Encapsulating Line Height is the Default Line Height.

## Phase 2: Determine Default Line Height by Font Metrics
- Reasoning:
	- This change will ensure that users who do not explicitly set a Line Height are by default using a value that is enough to make sure the Reported Text Height encapsulates all characters appropriately.
[...]

PR #201, opened 2024-03-14
Text Autosize Behavior

### LINKED ISSUE #28: Add w/h to textnode

frank-weindel:
Depending on the set contain mode, after the text 'loaded' event is emitted,
the text node may have either its width and height updated
to match the rendered size of the text.

When contain mode is 'none', both the width and height
properties are updated.

When contain mode is 'width', only the height property is updated.

When contain mode is 'both', neither property is updated.

Resolves #28


chiefcll (approved):
LGTM

pecoram:
Is it correct that letters such as "g" are cut off in snapshots ?
Granted that I haven't done any tests, with this modification does the alignCenter still work?

frank-weindel:
The descenders being cut of from the reported text height is existing behavior. I've got a proposal to fix that and make the reporting more consistent:
- #170 

As for #183, are you 100% sure? I wouldn't expect much of any behavior change when contain is set to "both". I will take your word for it though.

scailbc:
Hello, I'm seeing an error on word wrap that could be related.
Screenshot 2024-03-15 at 17 23 19
Screenshot 2024-03-15 at 17 23 42
As you can see a word with a single character is not wrapped correctly, while with 2 characters works fine.

Here is the code:
```js
<View x={400} y={500} width={100} height={100} color={0xff0000ff}>
    <Text fontSize={20} width={100} contain="width">
        Lorem ips e dolor
    </Text>
</View>
```

I'm using the renderer v0.7.5. Can you give it a check?

wouterlucas (approved):
LGTM

PR #203, opened 2024-03-18
ImageTexture: Fall back to HTMLImageElement in case createImageBitmap is not available

wouterlucas:
In case the browser does not support createImageBitmap the renderer will fallback to using the Image() constructor using HTMLImageElement to fetch images.

**Note** This is slower over using createImageBitmap, Lightning will throw a console.warn to warn its users. 

Addresses: https://github.com/lightning-js/renderer/issues/192 and follows the same fallback pattern as Lightning 2.
This feature will allow the renderer to work on Chrome < 50 where createImageBitmap is not supported. See https://caniuse.com/createimagebitmap for more information.

### DIFF
--- a/src/core/CoreTextureManager.ts
+++ b/src/core/CoreTextureManager.ts
@@ -145,7 +145,9 @@ export class CoreTextureManager {
     Texture,
     { cacheKey: string | false; count: number }
   > = new WeakMap();
-  imageWorkerManager: ImageWorkerManager;
+  imageWorkerManager: ImageWorkerManager | null = null;
+  hasCreateImageBitmap = !!self.createImageBitmap;
+  hasWorker = !!self.Worker;
   /**
    * Renderer that this texture manager is associated with
    *
@@ -157,7 +159,16 @@ export class CoreTextureManager {
 
   constructor(numImageWorkers: number) {
     // Register default known texture types
-    this.imageWorkerManager = new ImageWorkerManager(numImageWorkers);
+    if (this.hasCreateImageBitmap && this.hasWorker) {
+      this.imageWorkerManager = new ImageWorkerManager(numImageWorkers);
+    }
+
+    if (!this.hasCreateImageBitmap) {
+      console.warn(
+        '[Lightning] createImageBitmap is not supported on this browser. ImageTexture will be slower.',
+      );
+    }
+
     this.registerTextureType('ImageTexture', ImageTexture);
     this.registerTextureType('ColorTexture', ColorTexture);
     this.registerTextureType('NoiseTexture', NoiseTexture);
--- a/src/core/renderers/webgl/WebGlCoreCtxTexture.ts
+++ b/src/core/renderers/webgl/WebGlCoreCtxTexture.ts
@@ -23,6 +23,7 @@ import type { WebGlContextWrapper } from '../../lib/WebGlContextWrapper.js';
 import type { Texture } from '../../textures/Texture.js';
 import { isPowerOfTwo } from '../../utils.js';
 import { CoreContextTexture } from '../CoreContextTexture.js';
+import { isHTMLImageElement } from './internal/RendererUtils.js';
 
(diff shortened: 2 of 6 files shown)


src/core/CoreTextureManager.ts
```diff
@@ -145,7 +145,9 @@ export class CoreTextureManager {
     Texture,
     { cacheKey: string | false; count: number }
   > = new WeakMap();
-  imageWorkerManager: ImageWorkerManager;
+  imageWorkerManager: ImageWorkerManager | null = null;
+  hasCreateImageBitMap = !!window.createImageBitmap;
```
frank-weindel, line 149: `self` should be used instead of `window` as `window` is not compatible with running Core on a worker. Also can you lower-case the M in `BitMap`?

src/core/renderers/webgl/WebGlCoreCtxTexture.ts
```diff
@@ -135,7 +135,8 @@ export class WebGlCoreCtxTexture extends CoreContextTexture {
     // upload any data to the GPU.
     if (
       textureData.data instanceof ImageBitmap ||
-      textureData.data instanceof ImageData
+      textureData.data instanceof ImageData ||
+      textureData.data instanceof HTMLImageElement
```
frank-weindel, line 139: Running on a worker, `HTMLImageElement` will be undefined and I believe would throw an error.

src/core/CoreTextureManager.ts
```diff
@@ -157,7 +159,16 @@ export class CoreTextureManager {
...
     // Register default known texture types
-    this.imageWorkerManager = new ImageWorkerManager(numImageWorkers);
+    if (this.hasCreateImageBitMap && this.hasWorker) {
+      this.imageWorkerManager = new ImageWorkerManager(numImageWorkers);
+    }
+
+    if (!this.hasCreateImageBitMap) {
+      console.warn(
+        '[Lightning] createImageBitmap is not supported on this browser. ImageTexture will be slower.',
+      );
```
michielvandergeest, line 169: Does it make sense to throw a warning for something that can't be addressed by a developer anyway? It simply depends on the support of the browser if the ImageTexture will be used instead of createImageBitmap, not a user defined setting or something.

src/core/CoreTextureManager.ts
wouterlucas: sure thing! Will update.

src/core/renderers/webgl/WebGlCoreCtxTexture.ts
wouterlucas: mmm okay will have to figure out a way to support both of the path, will have a look tomorrow.

src/core/CoreTextureManager.ts
wouterlucas: It's what Lightning 2 also did. And yes I do think it makes sense, the only one who will see this is the developer and because it is something that falls back automatically the developer might not necessarily be aware of what's happening. Since it will affect performance, best case it saves the dev some time debugging why images are loading slower on a particular device versus one that does support `createImageBitmap`.

wouterlucas:
Validated with `driver=threadx`, straightened out a left over `window` reference in the `ImageWorker`. With that works like a charm.

frank-weindel:
Just for confirmation was this tested and works on Chrome < 50?

src/core/renderers/webgl/internal/RendererUtils.ts
```diff
@@ -129,3 +129,20 @@ export function createIndexBuffer(glw: WebGlContextWrapper, size: number) {
...
+
+/**
+ * Checks if an object is of type HTMLImageElement.
+ * This is used because we cant check for HTMLImageElement directly when the
+ * renderer is running in a seperate web worker context.
+ *
+ * @param obj
+ * @returns
+ */
+export function isHTMLImageElement(obj: unknown): obj is HTMLImageElement {
```
frank-weindel, line 141: Nice.

frank-weindel (approved):

PR #205, opened 2024-03-20
FPS: Update FPS when in idle RAF

wouterlucas:
Currently we do not fire FPS update events when the renderloop is in RAF idle state. This can give the perception that the app is stuck on the previously rendered frame speed.

This changes the FPS events to fire on RAF idle loops as well.

### DIFF
--- a/src/core/Stage.ts
+++ b/src/core/Stage.ts
@@ -243,11 +243,15 @@ export class Stage extends EventEmitter {
 
     renderer?.render();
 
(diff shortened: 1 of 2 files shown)


frank-weindel (approved):

michielvandergeest (approved):
LGTM

Tested with the FPS counter in Blits and this seems to give more realistic / expected results. 👍

PR #206, opened 2024-03-20
Added undefined check to text autosize behavior.

michielvandergeest:
Adding a check for undefined width or height on text node fixes a case where text is not properly displayed in Blits upon first app load (introduced by the changes in #201)

### DIFF
--- a/src/core/CoreTextNode.ts
+++ b/src/core/CoreTextNode.ts
@@ -131,6 +131,9 @@ export class CoreTextNode extends CoreNode implements ICoreTextNode {
   }
 
   override set width(value: number) {
+    if (value === undefined) {
+      return;
+    }
     this.props.width = value;
     this.textRenderer.set.width(this.trState, value);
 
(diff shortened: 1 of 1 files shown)


wouterlucas (approved):
LGTM 👍

frank-weindel:
I'm reluctant to accept this as it should be enforced by the specified type and there are quite a number of places where runtime checks like this could go to protect against errant JavaScript usage. But if you can add comments describing why it's needed (and promise it won't open a can of worms 🪱 ) I could make an exception.

Even when not using TypeScript the types should be regarded as part of the API spec. And not conforming properly to the API spec can yield undefined results.

chiefcll:
Why is the width and height possibly being set to undefined? Adding this would mean you should add an undefined check to every property. But you're really protecting developers from saying `node.x = undefined` - which should throw an error.

michielvandergeest:
No need for exceptions at the risk of opening a can worms! Fixed it another way around, so closing this.

Issue #196, opened 2024-03-12
emit onIdle event


chiefcll:
For folks who care about performance - I'd like the renderer to offer an onIdle event that I can listen for that is fired once when the requestAnimation doesn't have work to do. It should fire every time it finishes processing work and becomes idle for the first time - this will allow application developers to have a simple hook to do logging when thread is idle.

PR #208, opened 2024-03-21
feat: #196 add idle event

chiefcll:
Emit an idle event when renderer isnt performing any actions. Useful for frameworks to do work when nothing is happening. Also threadx doesnt need this as its a separate thread.

### DIFF
--- a/src/main-api/ICoreDriver.ts
+++ b/src/main-api/ICoreDriver.ts
@@ -63,4 +63,6 @@ export interface ICoreDriver {
   onFpsUpdate(fpsData: FpsUpdatePayload): void;
 
   onFrameTick(frameTickData: FrameTickPayload): void;
+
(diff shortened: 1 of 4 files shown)


src/main-api/ICoreDriver.ts
```diff
@@ -63,4 +63,6 @@ export interface ICoreDriver {
   onFpsUpdate(fpsData: FpsUpdatePayload): void;
 
   onFrameTick(frameTickData: FrameTickPayload): void;
+
+  onIdle?(): void;
```
frank-weindel, line 67: Don't think the optional is required here.

chiefcll: It is otherwise ThreadX needs it as well

frank-weindel (approved):

PR #210, opened 2024-03-21
Perf: Remove unnecessary cascading of premultiply colors update

frank-weindel:
dependency-graph
According to the dependency graph for CoreNodes, and the logic as I see it in the code, an update to a parent's premultiplied colors should not need to propagate/cascade down to its children.


wouterlucas (approved):
LGTM

Issue #159, opened 2024-02-06
Linear gradients performance


wouterlucas:
Currently the linear gradient doesn't perform, this needs to be investigated.
Need to check if radial gradients have the same issue.

toiladoong:
I have the same problem on TV, do you have a solution?

chiefcll:
Yes - we removed it an used a png
https://github.com/lightning-js/solid-demo-app/blob/main/src/components/Background.tsx#L61

chiefcll:
This can be closed out - @jfboeve - looks good thanks!

PR #214, opened 2024-03-26
Optimized gradient shaders

jfboeve:
Optimized gradient shaders for #159,

Moved color stops calculations to resolveDefaults function, since the calculation only needs to be done once in order to generate the shader source.

And removed the for loop in the fragment source.


frank-weindel (approved):

toiladoong:
Does this update resolve the issue of linear gradients not being displayed on TV?

jfboeve:
Not sure... which TV? any special environment? Does the radial gradient work by chance?

Issue #118, opened 2024-01-08
Add Hole Shader


chiefcll:
https://lightningjs.io/examples/#/shaders/hole

It's needed for video playback with picture in picture.

PR #215, opened 2024-03-26
added holepunch effect

jfboeve:
Added a holepunch effect for the dynamic shader, works inline with the Hole shader used in lightning 2, with an extra addition that the radius now supports an array similar to the radius effect.

Related issue:
#118


frank-weindel:
Please add a visual regression test if you can

chiefcll:
NIce..!

frank-weindel (approved):

PR #216, opened 2024-03-26
Fixed casing of out of viewport event.

michielvandergeest:
Event emitted for "out of viewport" was spelled with a capital `P` in `viewport`.

For consistency (and to match the provided example) this should probably have been a lowercase `p`.


wouterlucas (approved):
I think we should go back to the Lightning 2 pattern of putting at least 24 `_` s in front of it. 
Or no this is a public value, 17 `_` s will do. 

Otherwise this looks good to me.

frank-weindel (approved):

PR #218, opened 2024-03-28
fix Render Bounds check

jfboeve:
Fixed an issue where the render bounds check is using the wrong clippingRect and there was a miscalculation setting up the strict bound, this caused the in / out of viewport event not being triggered properly.


frank-weindel:
Nice. LGTM!

PR #219, opened 2024-03-28
added canBatchShaderProps to dynamic shader

jfboeve:
Got the batch check for the dynamic shader working properly reducing the draw calls:

Draws without batch check:
without_batch_check

Draws with batch check:
with_batch_check


chiefcll:
Is this for all dynamic shaders? Radius, borders etc? This is cool - excited to try out FPS

wouterlucas (approved):
LGTM!

PR #226, opened 2024-04-02
Node Texture Autosize Mode

### LINKED ISSUE #34: Add width / height to ImageTexture

frank-weindel:
When enabled, when a texture is loaded into the Node, the Node will automatically resize to the dimensions of the texture.

Text Nodes are always autosized based on their text content regardless of this mode setting.

Autosize Mode is false by default and must be explicitly set to `true` to be enabled.

_Based on previous work by @pecoram in #68_ 

Resolves #34


frank-weindel:
> Autosize Mode is false by default and must be explicitly set to `true` to be enabled.

Letting this line be up to debate. I wanted to rethink this and make it `true` by default since that is now essentially the mandatory behavior for text. But I feel that would break things a lot more.

I've decided to abandon having this be set to `true` when using the `src` property. If anyone disagrees with that let me know.

wouterlucas (approved):
LGTM!

chiefcll:
I think thats fine - it's easy to pass through that prop. Most of the time users will set the image size - and then they don't know the size they can set the flag.

chiefcll (approved):

PR #227, opened 2024-04-02
resolve isseus skipping inBounds state

jfboeve:
This fix resolves skipping the InBounds state when the renderMargin is not configured. In most cases it means that the size of the bounds check is the same as the size of the viewport. With this fix it will still fire an inBounds event when going directly from outOfBounds to inViewport and vice versa.


frank-weindel:
LGTM!

frank-weindel (approved):

PR #228, opened 2024-04-02
CoreNode: Set isRenderable to false if worldAlpha is 0

frank-weindel:
In #211, the Render State and IsRenderable result calculations were made into distinct update stages. This enables us to now use the World Alpha result when deciding if a node is renderable or not. Meaning if a parent's `alpha` is set to 0, neither itself nor any of its children will be added to the rendering pipeline.


erikhaandrikman (approved):

wouterlucas (approved):
LGTM

jfboeve (approved):

Issue #236, opened 2024-04-06
Feature: DOM inspector "extras"


elsassph:
DOM inspector is great for development, but insufficient for test automation.

Test automation needs to allow setting extra `dataset` fields, typically to provide a `testid` and mark the `focused` element.

wouterlucas:
anything you set as a data field will be exposed as data-<key> attribute by the inspector. 

You can find it here:
https://github.com/lightning-js/renderer/blob/main/src/main-api/INode.ts#L423

elsassph:
Awesome @wouterlucas

It's **mostly** working: one issue is that it doesn't seem possible to clear a data attribute.

Our automation uses `[data-focused]` to identify the focused element. Setting `data = { focused: '' }` can set the value, but I can't "unset" it.

elsassph:
Also ideally it shouldn't set a field which has `undefined` value

PR #237, opened 2024-04-08
Clear data attribute when value is undefined

elsassph:
Fixes issue in #236 comments: unable to remove an attribute.

When using `data` prop:

1. Once a value is set, it can't be removed; it can only be replaced by another value,
2. Passing an `undefined` value results in `"undefined"` being serialized as the dataset value.

I thing there is an opportunity to use an `undefined` value to clear a property.

Note: I wasn't able to run visual regression tests and had issues with pre-commit hooks.


wouterlucas (approved):
LGTM! Thanks!

chiefcll (approved):

frank-weindel (approved):

Issue #231, opened 2024-04-05
Viewport events - inBound event fired right before going outOfBounds


chiefcll:
When an item goes outOfViewport it also triggers inBounds and outOfBounds at same time.

Expectation is outOfViewport & outOfBounds would get triggered only.

wouterlucas:
Solved in #296

PR #238, opened 2024-04-08
fix viewport events

jfboeve:
I was a bit over eager with the events, it does indeed make more sense to not fire the inBounds again when going out of viewport.

Related to:
#231


chiefcll (approved):

jfboeve:
Closing this PR since we have a better working solution on the dev branch

PR #241, opened 2024-04-08
Added production renderer setting for specifying environment

michielvandergeest:
Currently the renderer depends on bundler specific `import.meta.env` for setting its production environment and skipping `assertTruthy` checks.

With this change, the production environment can be set from the App via a renderer setting. In an app the environment can still rely on `import.meta.env` (or any other mechanism to set the environment), but this leaves the renderer free of any bundler specific logic.

This solves issues when importing the renderer as a standalone prebuilt package (as is the case in the Lightning playground).

Open to suggestions on the naming. I have considered to name the property `environment` and allowing different strings (i.e. production, staging, testing, development), but ended up keeping it simple as it only seems to affect the assertTruthy check when running in prod.

### DIFF
--- a/src/main-api/Inspector.ts
+++ b/src/main-api/Inspector.ts
@@ -153,7 +153,8 @@ export class Inspector {
   private scaleY = 1;
 
   constructor(canvas: HTMLCanvasElement, settings: RendererMainSettings) {
-    if (isProductionEnvironment()) return;
+    if (isProductionEnvironment() || (import.meta.env && import.meta.env.PROD))
+      return;
 
     if (!settings) {
       throw new Error('settings is required');
--- a/src/env.d.ts
+++ b/src/env.d.ts
@@ -1,7 +1,0 @@
-interface ImportMetaEnv {
-  readonly PROD: boolean;
-}
-
-interface ImportMeta {
-  readonly env: ImportMetaEnv;
-}
--- a/src/env.d.ts
+++ b/src/env.d.ts
@@ -1,0 +1,7 @@
+interface ImportMetaEnv {
+  readonly PROD: boolean;
+}
+
+interface ImportMeta {
+  readonly env: ImportMetaEnv;
+}
--- a/src/main-api/RendererMain.ts
+++ b/src/main-api/RendererMain.ts
@@ -275,7 +275,8 @@ export interface RendererMainSettings {
    * Production environment
    *
    * @remarks
-   * Sets the renderer to run in production mode
+   * Sets the renderer to run in production mode as an override variable for cases the `import.meta.env.PROD`
+   * environment variable isn't available
    *
    * When set to `true` assertTruthy checks are skipped and ensures that
    * the inspector will not be attached, regardless of the `enableInspector` setting
@@ -432,7 +433,9 @@ export class RendererMain extends EventEmitter {
 
     targetEl.appendChild(canvas);
 
-    if (enableInspector && !isProductionEnvironment()) {
+    if (isProductionEnvironment() || (import.meta.env && import.meta.env.PROD))
+      return;
+    if (enableInspector) {
       this.inspector = new Inspector(canvas, resolvedSettings);
     }
   }
--- a/src/main-api/RendererMain.ts
+++ b/src/main-api/RendererMain.ts
@@ -40,7 +40,7 @@ import type { TextureUsageTracker } from './texture-usage-trackers/TextureUsageT
 import { EventEmitter } from '../common/EventEmitter.js';
 import { Inspector } from './Inspector.js';
 import { santizeCustomDataMap } from '../render-drivers/utils.js';
-import { isProductionEnvironment } from '../utils.js';
+import { isProductionEnvironment, setProductionEnvironment } from '../utils.js';
 
 /**
  * An immutable reference to a specific Texture type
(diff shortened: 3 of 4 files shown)


chiefcll:
This needs to be done via a build flag otherwise Vite won't be able to treeshake any of the code out. It will be included in production which will slow down production apps.

michielvandergeest:
I don't think that's true. Running a `vite build` with the current main, the final bundle _also_ contains the `assertTruthy` function and several functions throughout the codebase calling it. 

That functionality got dropped in https://github.com/lightning-js/renderer/pull/212

michielvandergeest:
If we care about treeshaking this function (which will probably get optimized by JIT in no time) we should revert #212 _and_ implement some proper dist builds, because it's currently breaking cases where the renderer is imported from a CDN without going through a bundler.

chiefcll:
Ah - was hoping #212 worked - I copied it.
```js
function isDevEnv(): boolean {
  return import.meta.env && import.meta.env.DEV;
}
export const isDev = isDevEnv() || false;
```

Vite replaces import.meta.env.DEV with an empty string or true - I haven't figured out a way to make it work in browser and also build. The problem is Vite can't tree shake anything that is determined at runtime, so we'd be sending down extra JS code. I have rollup doing 
```js
plugins: [
      replace({
        'import.meta.env.MODE': '"production"',
        preventAssignment: true,
      }),
    ],
   ```
   
   Might need to spend some time doing a build for CDN vs build for apps

**Comment by michielvandergeest** [comment]
Tree shaking in general is important yes, the less code and less function invocations the better. But I think you are overestimating the impact of this specific case.

Anyway, for the sake of moving this along, there’s a way to do best of both worlds, that actually reenables the tree shaking functionality that got removed in #212 (since bundlers aren't capable of that level of static analysis). And at the same time it keeps the setting to force production at runtime for non-bundler use cases.

**Comment by elsassph** [comment]
Using functions (like `isProductionEnvironment()` or `isDevEnv()`) to mark code to skip like that won't do anything to code tree shaking / code size.

Minifiers only remove code when expressions are constants like:
```js
if (DEBUG) {
  doSomething();
}
```
where `DEBUG` is a value injected by the bundler (Webpack `definePlugin` or Vite `define` object).

With such logic, code can be actually removed from production builds.

PS: inspector is something that needs to be possible to enable in production builds! It's important for test automation.

src/main-api/Inspector.ts
```diff
@@ -153,7 +153,8 @@ export class Inspector {
   private scaleY = 1;
 
   constructor(canvas: HTMLCanvasElement, settings: RendererMainSettings) {
-    if (isProductionEnvironment()) return;
+    if (isProductionEnvironment() || (import.meta.env && import.meta.env.PROD))
```
elsassph, line 156: We need to be able to enable the inspector in production builds for test automation.

elsassph:
As a side note, it would be better to avoid core libraries using bundler-specific logic. It's looking a bit messy.

elsassph:
More about that: https://blog.battlefy.com/removing-assertions-from-production-build-with-vite

michielvandergeest:
> Using functions (like `isProductionEnvironment()` or `isDevEnv()`) to mark code to skip like that won't do anything to code tree shaking / code size.

Yes, aware of that. See also my previous comment. I didn't consider this initially as it was already not being treeshaken. My latest commit reenables the treeshaking, while keeping an override for other non-bundler use cases.

Not ideal, but it solves a current need.

> 
> Minifiers only remove code when expressions are constants like:
> 
> ```js
> if (DEBUG) {
>   doSomething();
> }
> ```

Exactly, bundlers / minifiers aren't capable of doing a full analysis and rely on these type of constant expression. The whole reason why PR 212 broke the tree shaking.

> PS: inspector is something that needs to be possible to enable in production builds! It's important for test automation.

I guess that makes sense to me. I would be fine with removing that check, but then would suggest to at least log a warning. I guess it does mean though that the inspector is always included and never treeshaken from production builds. @wouterlucas what do you think?

elsassph:
My point is that this code is totally wrong:
```js
    if (isProductionEnvironment() || (import.meta.env && import.meta.env.PROD))
```
you should stick to `if (import.meta.env.PROD)` as it can be supported by all bundlers and will be effectively removed during production build.

elsassph:
Sorry I didn't notice it was about non-bundler cases, but really non-bundler cases should be using a production build of the library.

michielvandergeest:
Yes, the key being "removed during production build".

For code that doesn't go through a bundler, but is run straight of a CDN, the `import.meta.env` object doesn't exist and will throw either an error or will run the assertion (which can also fail in some cases).

This is a (less than ideal, as already admitted) workaround to solve a current need we have. More than happy to refactor / rethink later on, but for now it's blocking things

elsassph:
Through a CDN you should be loading the production build of the library so I don't think this is a use case.

frank-weindel:
> More about that: https://blog.battlefy.com/removing-assertions-from-production-build-with-vite because you are misusing the `import.meta.env.PROD` variables.

@elsassph I'm pretty sure this was the article I referenced when I originally got this working. I see it indeed broke in #212 likely because of the `import.meta.env` check in the `isProductionEnvironment()` function. Vite only replaces the `import.meta.env.PROD` to `true`/`false` prior to the tree shaking. It doesn't touch the `import.meta.env` making the function still potentially alive.

@michielvandergeest @jfboeve We may very well want a special build that is built for direct importing on the web. It could, for instance, have `import.meta.env.PROD` be replaced with `window.PROD`/`self.PROD` or something like that. And also solve the other issue around `instanceof` by having all exports from the Main API and Core API exported together.

I see this use case as special, and not something we should ever encourage anyone to use outside of the context of a web playground. We're relying on the bundler for some core things from replacing this PROD variable to properly handling the web worker and Core/Main API separation/deduplication for Core Extensions that enables multi-theaded mode to be easy to configure and switch to.

elsassph:
What again was the point of `isProductionEnvironment()` again?

Only inline use of import meta will create dead code which can be eliminated:
```js
if (false) {
   // This is now dead code
}
```

PR #242, opened 2024-04-10
Fix: Race condition on this.StoppedPromise with animations

wouterlucas:
This PR moves the creation of the promise to 1 spot for both the `start()` case as well as the `loop` case, whereas `waitUntilStopped()` only returns what's created prior. Avoiding potential race conditions.

In our tests we see an animation that gets created early on during boot of the app get "stuck", never getting a promise resolve and essentially hanging the calling party.

 *The theory*:
`waitUntilStopped()` calls  `this.makeStoppedPromise();` and grabs a promise from `this.stoppedPromise`. However if `this.makeStoppedPromise();` 's creation of its `promise` is pushed to the next tick because there is a lot going on (e.g. an app is booting up) essentially not assigning `this.stoppedResolve` yet this could lead to an unresolved promise (with a short animation).

 *AND/OR*:

Both the animation `start()` AND `waitUntilStopped` are calling `this.makeStoppedPromise();` nearly simulatenously or across tick boundaries, resulting in both passing the `(this.startedResolve === null)` check and essentially overwriting each others `this.startedResolve` values. Causing the first `this.stoppedResolve` to never resolve.

The following is an app loaded on Glass with the bug:
StuckAnimation

By adding a console log in the animation start, you can see its stuck per:
```js
diff --git a/src/core/animations/CoreAnimation.ts b/src/core/animations/CoreAnimation.ts
index 251b048..f27d0dd 100644
--- a/src/core/animations/CoreAnimation.ts
+++ b/src/core/animations/CoreAnimation.ts
@@ -114,6 +114,7 @@ export class CoreAnimation extends EventEmitter {

     if (this.delayFor <= 0 && this.progress === 0) {
       this.emit('start', {});
+      console.log('Starting animation', this.node.id, this.propsList);
     }

     this.progress += dt / duration;
```

Results in:
StuckAnimation2

With the suggested change the animation gets unloaded and CPU load normalizes:
StuckAnimationFix

And no longer prints a stuck animation in the console.log:
StuckAnimationFix2

### DIFF
--- a/src/core/animations/CoreAnimationController.ts
+++ b/src/core/animations/CoreAnimationController.ts
@@ -48,10 +48,16 @@ export class CoreAnimationController implements IAnimationController {
   state: AnimationControllerState;
 
   start(): IAnimationController {
-    this.makeStartedPromise();
+    this.startedPromise = new Promise((resolve) => {
+      this.startedResolve = resolve;
+    });
+
     this.animation.once('start', this.started.bind(this));
 
-    this.makeStoppedPromise();
+    this.stoppedPromise = new Promise((resolve) => {
+      this.stoppedResolve = resolve;
+    });
+
     this.animation.once('finished', this.finished.bind(this));
 
     // prevent registering the same animation twice
@@ -87,33 +93,13 @@ export class CoreAnimationController implements IAnimationController {
   }
 
   waitUntilStarted(): Promise<void> {
-    this.makeStartedPromise();
-    const promise = this.startedPromise;
-    assertTruthy(promise);
-    return promise;
+    assertTruthy(this.startedPromise);
+    return this.startedPromise;
   }
 
   waitUntilStopped(): Promise<void> {
-    this.makeStoppedPromise();
-    const promise = this.stoppedPromise;
-    assertTruthy(promise);
-    return promise;
-  }
-
-  private makeStartedPromise(): void {
-    if (this.startedResolve === null) {
-      this.startedPromise = new Promise((resolve) => {
-        this.startedResolve = resolve;
-      });
-    }
-  }
-
-  private makeStoppedPromise(): void {
-    if (this.stoppedResolve === null) {
-      this.stoppedPromise = new Promise((resolve) => {
-        this.stoppedResolve = resolve;
-      });
-    }
+    assertTruthy(this.stoppedPromise);
+    return this.stoppedPromise;
   }
 
   private started(): void {
@@ -139,6 +125,11 @@ export class CoreAnimationController implements IAnimationController {
     this.stoppedResolve = null;
 
     if (loop) {
+      // generate new promise and resolve function
+      this.stoppedPromise = new Promise((resolve) => {
+        this.stoppedResolve = resolve;
+      });
+
       return;
     }
 


michielvandergeest (approved):
LGTM. Tested and this fix works.

I have not seen cases where the `this.startedResolve` didn't get fully executed, but I guess it makes sense to remove the nullifying there for the same reasons.

src/core/animations/CoreAnimationController.ts
```diff
@@ -87,33 +93,13 @@ export class CoreAnimationController implements IAnimationController {
   }
 
   waitUntilStarted(): Promise<void> {
-    this.makeStartedPromise();
-    const promise = this.startedPromise;
-    assertTruthy(promise);
-    return promise;
+    assertTruthy(this.startedPromise);
+    return this.startedPromise;
```
frank-weindel, line 97: How can you waitUntilStarted() if the `startedPromise` isn't created until you start()? 👀

frank-weindel (changes requested):
See comment

src/core/animations/CoreAnimationController.ts
wouterlucas: Race conditions can be funny like that 🥲 

What happens, in this case blits does:
```
        f.start()
          .waitUntilStarted()
          .then((animation) => {
            // fire transition start callback if specified
            transition.start &&
              typeof transition.start === 'function' &&
              transition.start.call(this.component, this, prop, startValue)
            // continue the chain
            animation
              .waitUntilStopped()
              .then(() => delete this.scheduledTransitions[prop])
              .then(() => {
                // fire transition end callback if specified
                transition.end &&
                  typeof transition.end === 'function' &&
                  transition.end.call(this.component, this, prop, this.node[prop])
              })
              .then(resolve)
```

Granted this can be optimized too, but for now lets go through what happens. Essentially:
1. Start an animation
2. And then `.waitUntilStarted()`

This never gets resolved, making the rest / cleanup of the animation never finish leaving the animation stuck and the router prevents the view from being destroyed
For reference this happens here: https://github.com/lightning-js/blits/blob/master/src/engines/L3/element.js#L356C11-L356C30

and has an effect here:
https://github.com/lightning-js/blits/blob/master/src/router/router.js#L248

This only happens when there is a lot going on, during the Splash screen. The app is loading, data is being fetched and the splash transition happens. If you look at the CPU graph you can see the system is briefly at 100% during this.

Now from what we've observed:
1. Promise is created on the renderer, this returns. Renderer initiates the creation of a resolved promise here: https://github.com/lightning-js/renderer/blob/main/src/core/animations/CoreAnimationController.ts#L51
2. This goes through a check to see if there already is a promise and this is null, per here: https://github.com/lightning-js/renderer/blob/main/src/core/animations/CoreAnimationController.ts#L104
3. Simultanously as a response to 1. Blits calls `waitUntilStarted()` this ALSO creates a resolve promise per: https://github.com/lightning-js/renderer/blob/main/src/core/animations/CoreAnimationController.ts#L90
4. This goes through the same check as 2. however because the load is high both this check and the one happening at 2 return false per: https://github.com/lightning-js/renderer/blob/main/src/core/animations/CoreAnimationController.ts#L104

It works fine if 2 and 4 happen in the same tick in sequence or with relative time in between. However due to the splash screen and app start there is a lot going on and 2 and 4 are executed across ticks prior to being finished resulting in a race condition where one overwrites the other.

What would fix this is create the resolve promise in 1 spot, at the start or at the loop point of the animation. Avoiding possible race conditions in the future. 

We could do a little more graceful checks if someone tries to call waitUntilStarted without actually starting the animation.

Let me know if we need to jump on a call to explain further.

frank-weindel: @wouterlucas I'm not questioning the issue.

It's interesting that Blits uses `waitUntilStarted()` that way but what about more general uses of it where `start()` gets called later?

Imagine `start()` is called after `waitUntilStarted()` (which to me seems a more natural use for it)
```
f.waitUntilStarted().then(() => {
  // Do something
});
// ^--- Throws an exception with this change

function onKeyPressHandler() {
  f.start();
}
```

michielvandergeest: In Blits the `waitUntilStarted`-promise is used to execute a callback when an animation is _actually_ starts, and not when it is initiated. This is to handle the case where there is a delay specified, which is handled in the renderer.

Previously the renderer didn't handle delay yet, so the delay was managed on the Blits side. Calling f.start() was then the exact moment the animation would start. That is no longer the case, hence this construction. 

The current code grew organically and admittedly is in need of  a good refactor 😂
But I think the general thought behind it (and the way the waitUntilStarted promise is used) still stands.

michielvandergeest: Adding to this:

If you think the API makes more sense when `waitUntilStarted()` is available on the result of `node.animate()` instead of returned by the `start()` function, then that's totally fine by me as well.

That construct doesn't sound wrong to me in any case, and happy to refactor around that API.

michielvandergeest: Also @wouterlucas

> And then .waitUntilStarted()
> This never gets resolved,

I'm not sure the problem ever occurred with `waitUntilStarted()` not resolving. I believe the issue was primarily related to the `waitUntilStopped()` being forever pending. That's how I first tracked down the issue by disabling the nullifying of the `waitUntilStopped`-promise

frank-weindel: Thanks for helping me understand the reasoning around `waitUntilStarted()`. It is a bit confusing from an intuitive perspective. Technically, even if there's a delay, the animation is starting when you issue `start()`. `waitUntilAfterDelay()` or something similar would be more intuitive. Regardless, it would be good to somehow fix this while keeping both of those promises available without preconditions (i.e. `start()` being called first)

frank-weindel: @wouterlucas 1, 2, 3, 4 should be happening completely synchronously. You can't get to 3 (the call to `waitUntilStarted()` without `start()` first returning. The promise constructor callbacks are executed synchronously as well. That would line up with what @michielvandergeest is saying.

wouterlucas: Hmm yes you are right, `waitUntilStopped` was the root symptom. Just moving anything in that function changes the timing which might give a false positive that the race condition is gone. I don't really see how removing that `null` changes the race condition (yet). The only place where I do see a race condition is this double `this.makeStartedPromise();` invocation. Might just mess up the scoping because we're potentially creating two independent Promises that creates a problem down the road when stopped needs to occur?

@frank-weindel they are not completely synchronous though, yes 1 and 3 are synchronous but 1 and 2 are not and neither is 3 and 4. Because we're calling `this.makeStartedPromise();` in two different places in short succession. It's not guaranteed that a new function invocation of `this.makeStartedPromise();` happens in the same tick, especially when the ticks are really very full due to the interpreter also doing a lot of other background stuff as well (loading/parsing data, spawning new stuff, creating components). 1 and 3 might happen in order, but both their invocations might be pushed to the a new tick because there is a lot of other stuff to do and then a race condition occurs. Whether or not that is the root cause of the stopped case, I don't know.

Other than that, I can't see any other place where a race condition might occur. Though more then happy to go through it step by step again and more closely look at the stopped case.

Regardless of constructing a different approach around the API, I think we should prevent this from happening.

frank-weindel:
Replaced by #274

PR #244, opened 2024-04-11
fix: cross image support on base images

chiefcll:
Gets images loading in older browsers when cross domain. This is the same as what L2 did. Tested with TMDB app and its working.

### DIFF
--- a/src/core/textures/ImageTexture.ts
+++ b/src/core/textures/ImageTexture.ts
@@ -117,8 +117,7 @@ export class ImageTexture extends Texture {
       };
     } else {
       const img = new Image();
-      if (!(src.substr(0, 5) == 'data:')) {
-        // Base64.
+      if (!(src.substr(0, 5) === 'data:')) {
         img.crossOrigin = 'Anonymous';
       }
       img.src = src;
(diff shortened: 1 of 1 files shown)


src/core/textures/ImageTexture.ts
```diff
@@ -117,6 +117,10 @@ export class ImageTexture extends Texture {
       };
     } else {
       const img = new Image();
+      if (!(src.substr(0, 5) == 'data:')) {
```
frank-weindel, line 120: Could you help my OCD by switching that to a `===`?

src/core/textures/ImageTexture.ts
```diff
@@ -117,6 +117,10 @@ export class ImageTexture extends Texture {
...
+        // Base64.
```
frank-weindel, line 121: What does this comment mean here? If we're in this block, then the src did not start with `data:` and hence is not likely Base64.

chiefcll: Yea - was in lng2 - they had a PS4 check too which we dont have... removed it

frank-weindel (approved):

PR #245, opened 2024-04-12
fix(ImageWorker): fix compatibility with older Tizen

el-gringo:
- allow ImageWorker to works without Worker
- add polyfills for queueMicrotask & Object.fromEntries


src/utils.ts
```diff
@@ -19,6 +19,13 @@
 
 import type { ContextSpy } from './core/lib/ContextSpy.js';
 
+let promise
+
+if (window.queueMicrotask === undefined) {
+  window.queueMicrotask = requestIdleCallback;
+}
```
chiefcll, line 26: I assume core-js will fix this... Though in Solid build process I'm ignoring renderer so it might not be optimizing to fix this. Not sure if we want this one piece in here cause don't know what will happen when renderer is optimized as well. I'll need to test it out.

src/utils.ts
```diff
@@ -19,6 +19,13 @@
 
 import type { ContextSpy } from './core/lib/ContextSpy.js';
 
+let promise
```
chiefcll, line 22: Should this be here?

wouterlucas:
Replacing with https://github.com/lightning-js/renderer/pull/305 as original contributor is MIA

PR #246, opened 2024-04-14
Canvas 2D renderer

elsassph:
Initial canvas 2D renderer.

Added `renderMode: 'webgl' | 'canvas'` renderer option.

https://github.com/lightning-js/renderer/assets/297963/5aaac948-7ae7-4c1c-a255-6de0a68359b8

Limitation:
- Only horizontal & vertical color gradients,
- Only `RoundedRectangle` "shader".

Future potential PRs:
- Canvas-text textures optimization to avoid doing Canvas -> ImageData -> Canvas
- Maybe a c2d shader API

Unsupported:
- All shaders (gracefully ignored) excepted `RoundedRectangle`,
- SDF text (fallback to canvas text).

Aside from that, all the examples render correctly! Rotation, clipping, tinting, spritesheets...

### DIFF
--- a/src/core/renderers/canvas/CanvasCoreRenderer.ts
+++ b/src/core/renderers/canvas/CanvasCoreRenderer.ts
@@ -24,7 +24,7 @@ import type { Texture } from "../../textures/Texture.js";
 import type { CoreContextTexture } from "../CoreContextTexture.js";
 import { CoreRenderer, type CoreRendererOptions, type QuadOptions } from "../CoreRenderer.js";
 import { CanvasCoreTexture } from "./CanvasCoreTexture.js";
-import { formatRgba, parseColor } from "./internal/ColorUtils.js";
+import { formatRgba, parseColor, type IParsedColor } from "./internal/ColorUtils.js";
 
 export class CanvasCoreRenderer extends CoreRenderer {
 
@@ -67,7 +67,7 @@ export class CanvasCoreRenderer extends CoreRenderer {
   addQuad(quad: QuadOptions): void {
     const ctx = this.context;
     const {
-      tx, ty, width, height, alpha, colorTl, ta, tb, tc, td, clippingRect
+      tx, ty, width, height, alpha, colorTl, colorTr, colorBr, ta, tb, tc, td, clippingRect
     } = quad;
     let texture = quad.texture;
     let ctxTexture: CanvasCoreTexture | undefined = undefined;
@@ -92,6 +92,7 @@ export class CanvasCoreRenderer extends CoreRenderer {
     const color = parseColor(colorTl);
     const hasTransform = ta !== 1;
     const hasClipping = clippingRect.width !== 0 && clippingRect.height !== 0;
+    const hasGradient = colorTl !== colorTr || colorTl !== colorBr;
 
     if (hasTransform || hasClipping) {
       ctx.save();
@@ -132,6 +133,26 @@ export class CanvasCoreRenderer extends CoreRenderer {
         ctx.drawImage(image, tx, ty, width, height);
       }
       ctx.globalAlpha = 1;
+    } else if (hasGradient) {
+      let endX: number = tx;
+      let endY: number = ty;
+      let endColor: IParsedColor;
+      if (colorTl === colorTr) {
+        // vertical
+        endX = tx;
+        endY = ty + height;
+        endColor = parseColor(colorBr);
+      } else {
+        // horizontal
+        endX = tx + width;
+        endY = ty;
+        endColor = parseColor(colorTr);
+      }
+      const gradient = ctx.createLinearGradient(tx, ty, endX, endY);
+      gradient.addColorStop(0, formatRgba(color));
+      gradient.addColorStop(1, formatRgba(endColor));
+      ctx.fillStyle = gradient;
+      ctx.fillRect(tx, ty, width, height);
     } else {
       ctx.fillStyle = formatRgba(color);
       ctx.fillRect(tx, ty, width, height);
--- a/src/core/renderers/canvas/CanvasCoreRenderer.ts
+++ b/src/core/renderers/canvas/CanvasCoreRenderer.ts
@@ -1,0 +1,141 @@
+/*
+ * If not stated otherwise in this file or this component's LICENSE file the
+ * following copyright and licenses apply:
+ *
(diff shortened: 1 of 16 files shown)


erikhaandrikman:
Wow this is great! Thanks for the contribution so far! We've been thinking about starting the Canvas2d renderer for Lightning 3, especially since some devices still require it and it's supported in Lightning 2, and now this drops out of the clear sky :)

Is there anything we can assist you with? 

**clearColor:** 
You are probably looking at `ctx.clearRect()` ?

**Gradients:** 
Besides linear, we could explore supporting radial gradients too:  https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/createRadialGradient

**Gounded rectangle:** 
Would probably be highly requested in a cavas2d renderer too, We could even explore rendering textures within a clipped region:  https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/clip

**Shaders:** 
Is indeed a harder topic since there is simply no support for it in Canvas2D context. However, we could investigate using filters as an alternative: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/filter

Really excited for this feature!

elsassph:
My goal for this initial PR will be to add at basic linear gradients. After that, I'd like to take the time to think about optimising text textures and maybe try rounded rectangle.

Realistically, it's not worth making the canvas 2D fallback as featured as the webgl one. LightningJS 2 did not support any webgl shader either.

erikhaandrikman:
Fair point, although L2 had fallback for GrayScale and BlurShader

elsassph:
Yes I think some form of c2d shader system can be eventually considered.

elsassph:
Renderer `clearColor` and simple horizontal/vertical gradients added. Now ready!

elsassph:
Alright, doing the `RoundedRectangle` was easy enough :D

frank-weindel:
Fantastic work @elsassph! Thanks for this amazing contribution and so skillfully finding a way to fit it neatly into the existing abstraction. I loaded this up and tested some examples and it all works super well.

I mostly have a few relatively minor comments which I'll leave inline on some changes.

One general comment I have is around tree-shakability. We don't necessarily need to solve this now, or even at all. But in the cases of apps that may opt not to use Canvas mode at all it would be good to decouple Canvas mode in such a way that it could be tree shaken off if it wasn't used. But maybe only if it gets much bigger in size. The additions seem fairly small right now.

elsassph:
Looking forward for comments.

I for one want more tree-shaking for sure. Thankfully canvas renderer is pretty small:
image

src/utils.ts
```diff
@@ -173,9 +183,10 @@ export function mergeColorAlphaPremultiplied(
   flipEndianess = false,
 ): number {
   const newAlpha = ((rgba & 0xff) / 255) * alpha;
-  const r = Math.trunc((rgba >>> 24) * newAlpha);
```
frank-weindel, line 176: Could you update the comment of this function to match its new behavior?

src/core/renderers/canvas/CanvasCoreTexture.ts
```diff
@@ -0,0 +1,137 @@
...
+    this.tintCache = undefined;
+    this.textureSource.setState('freed');
+    this.memManager.setTextureMemUse(this, 0);
+  }
+
+  updateMemSize(): void {
+    // Counting memory usage for:
+    // - main image
+    // - tinted image
+    // - should we count the source texture in addition?
```
frank-weindel, line 55: I don't think so. Unless the "main image" itself is holding onto the reference to the texture source's data, that data should be completely unreferenced and up for garbage collection.

elsassph: If the `textureSource` itself does not keep reference to the loaded data then indeed it should be ok.

src/core/renderers/canvas/CanvasCoreRenderer.ts
```diff
@@ -0,0 +1,177 @@
...
+    const { canvas, pixelRatio, clearColor } = options;
+    this.canvas = canvas as HTMLCanvasElement;
+    this.context = canvas.getContext('2d') as CanvasRenderingContext2D;
+    this.pixelRatio = pixelRatio;
+    this.clearColor = clearColor ? getRgbaComponents(clearColor) : undefined;
+  }
+
+  reset(): void {
+    // quick reset canvas
+    this.canvas.width = this.canvas.width ?? 1920;
```
frank-weindel, line 52: 1920?

src/core/text-rendering/renderers/SdfTextRenderer/SdfTextRenderer.ts
```diff
@@ -580,7 +581,8 @@ export class SdfTextRenderer extends TextRenderer<SdfTextRendererState> {
       return;
     }
 
-    const { renderer } = this.stage;
+    const renderer: WebGlCoreRenderer = this.stage.renderer as WebGlCoreRenderer;
+    assertTruthy(renderer.mode === 'webgl');
```
frank-weindel, line 585: I think an `instanceof` check would be cleanest here

src/core/renderers/canvas/CanvasCoreRenderer.ts
elsassph: Default resolution for full-hd, only for the sake of the `width` attribute being possibly `null`...

elsassph:
For tree shaking, ideally the Stage init would be async and the renderer loaded using a dynamic import.

src/core/renderers/canvas/CanvasCoreRenderer.ts
frank-weindel: From what I see it's not possible for it to be null. Though I do see if you do get rid of the null-coalescing operator it gives a `no-self-assign` eslint error, which you could safely inline ignore.

src/utils.ts
elsassph: Better?

frank-weindel: Yeah, if you can just adjust the NOTE line like below it will be good enough.
```
NOTE: Depending on the mode set by {@link setPremultiplyMode}, this method can return a PREMULTIPLIED alpha color which is generally only useful...
```

frank-weindel (approved):
LGTM!

PR #249, opened 2024-04-15
Fix: Optimize Get Timing Function for animations

wouterlucas:
This PR properly optimizes the getTimingFunction, which is called during animations, and drops the memize dependency. Applying memoization negatively impacts the JIT's ability to optimize function paths in named easing cases while it helps for custom timings. The suggested change improves performance for both named as well as custom timing cases.

There are 2 distinct cases for `getTimingFunction`: using named easing timings and using custom `cubic-bezier` timings. In the case of named timings the original function proved to be faster then the updated memized function, as the JIT was able to optimize the old switch case fairly well. 

The original performance hit came from dealing with `cubic-bezier`'s due to their required manual string operations and `parseFloat` to convert the string to a `number`. This change caches where needed, storing the `cubic-bezier`s in a key/value object and relies on a `regex` to parse the string. 

Measurements with the original implementation, the memized implementation and the optimized implementation in this PR:

With 1 named value:
```
[Log] original x 4,628 ops/sec ±0.78% (47 runs sampled) (index.ts, line 17)
[Log] memoized x 2,096 ops/sec ±0.50% (47 runs sampled) (index.ts, line 17)
[Log] optimized x 5,666 ops/sec ±0.69% (47 runs sampled) (index.ts, line 17)
```

The optimized is doing about 1.5-2x of the memoized implementation. 
With 1 named value the original implementation is pretty close to the optimized one due to JITs ability to graph the early path out of the switch and apply caching.

With 2 named values:
```
[Log] original x 2,755 ops/sec ±0.68% (45 runs sampled) (index.ts, line 17)
[Log] memoized x 1,410 ops/sec ±1.10% (46 runs sampled) (index.ts, line 17)
[Log] optimized x 4,907 ops/sec ±1.05% (46 runs sampled) (index.ts, line 17)
```

With 2 the difference is still pretty similar, though the original function starts to fall behind.

With 3 named values: 
```
[Log] original x 1,319 ops/sec ±5.07% (39 runs sampled) (index.ts, line 17)
[Log] memoized x 1,264 ops/sec ±0.84% (47 runs sampled) (index.ts, line 17)
[Log] optimized x 4,927 ops/sec ±0.58% (26 runs sampled) (index.ts, line 17)
```

The optimized function is rather unaffected by the named exports and now does 3-4x of both other implementations.
The interesting part here is that the original starts to match the memoized implementation and the hash map is starting to pay off.

However if we run the same test with just 1 unique  `cubic-bezier` it starts to get really interesting.

With 1 cubic-bezier:
```
[Log] original x 176 ops/sec ±4.78% (37 runs sampled) (index.ts, line 17)
[Log] memoized x 2,054 ops/sec ±0.45% (47 runs sampled) (index.ts, line 17)
[Log] optimized x 5,722 ops/sec ±0.69% (46 runs sampled) (index.ts, line 17)
```

Here is where the original implementation drops on the floor, this is due to the string splits and `parseFloat`. Memoize starts to shine because it can cache the response pretty well. 

Though the optimized is still at 2.5x+ over memoized. 

Whether we use 1 or 6 unique  `cubic-bezier`'s, the original remains at the floor:

With 6 unique cubic beziers:
```
[Log] original x 167 ops/sec ±5.40% (38 runs sampled) (index.ts, line 17)
[Log] memoized x 1,113 ops/sec ±2.06% (45 runs sampled) (index.ts, line 17)
[Log] optimized x 4,676 ops/sec ±2.01% (48 runs sampled) (index.ts, line 17)
```

Even the memoized starts to almost drop by 50% - the original function remains almost unchanged. This is the JIT still  being bogged down by the same `parseFloat` bottleneck.

The new implementation still sits pretty stable and now doing 4x+ over memoized. 
If we mix named and cubic, you get this (this was a suggested "real use case") you get the following.

With 2 named and 1 unique cubic bezier:
```
[Log] original x 470 ops/sec ±4.24% (44 runs sampled) (index.ts, line 17)
[Log] memoized x 1,294 ops/sec ±0.78% (47 runs sampled) (index.ts, line 17)
[Log] optimized x 4,844 ops/sec ±1.24% (47 runs sampled) (index.ts, line 17)
```

What's interesting to see is that the original function starts to pick up again, this is because for the 2 named timings the JIT can apply inline caching and apply DFG graphing through the switch case routes. 

The new implementation is rather unchanged, still sitting at a 4x over memoized.

Now a worst case test, all possible named timings and a load of cubic's. 

With 17 named and 30 unique cubic beziers:
```
[Log] original x 246 ops/sec ±7.62% (38 runs sampled) (index.ts, line 17)
[Log] memoized x 245 ops/sec ±1.39% (46 runs sampled) (index.ts, line 17)
[Log] optimized x 3,016 ops/sec ±1.50% (46 runs sampled) (index.ts, line 17)
```

This is where it starts to get funny, both the original and the memoized approach are on the floor. While the optimized function still sits at 3k ops, doing 12x of both. 

Tests run by adding 1000 animation easing requests and measuring performance using `benchmark.js` from jsperf.
Tested on WPEWebKit 2.38 running on a Glass TV. 

_small note_ this will likely even get better once we get FTL and OMG on WPEWebKit in late 2024 😎

### DIFF
--- a/src/core/utils.ts
+++ b/src/core/utils.ts
@@ -123,43 +123,41 @@ interface TimingFunctionMap {
   [key: string]: (time: number) => number | undefined;
 }
 
-const timingMapping: TimingFunctionMap = {
-  linear: (time: number): number => {
-    return time;
-  },
-  ease: getTimingBezier(0.25, 0.1, 0.25, 1.0),
-  'ease-in': getTimingBezier(0.42, 0, 1.0, 1.0),
-  'ease-out': getTimingBezier(0, 0, 0.58, 1.0),
-  'ease-in-out': getTimingBezier(0.42, 0, 0.58, 1.0),
-  'ease-in-sine': getTimingBezier(0.12, 0, 0.39, 0),
-  'ease-out-sine': getTimingBezier(0.12, 0, 0.39, 0),
-  'ease-in-out-sine': getTimingBezier(0.37, 0, 0.63, 1),
-  'ease-in-cubic': getTimingBezier(0.32, 0, 0.67, 0),
-  'ease-out-cubic': getTimingBezier(0.33, 1, 0.68, 1),
-  'ease-in-out-cubic': getTimingBezier(0.65, 0, 0.35, 1),
-  'ease-in-circ': getTimingBezier(0.55, 0, 1, 0.45),
-  'ease-out-circ': getTimingBezier(0, 0.55, 0.45, 1),
-  'ease-in-out-circ': getTimingBezier(0.85, 0, 0.15, 1),
-  'ease-in-back': getTimingBezier(0.36, 0, 0.66, -0.56),
-  'ease-out-back': getTimingBezier(0.34, 1.56, 0.64, 1),
-  'ease-in-out-back': getTimingBezier(0.68, -0.6, 0.32, 1.6),
-  'step-start': () => {
-    return 1;
-  },
-  'step-end': (time: number): number => {
-    return time === 1 ? 1 : 0;
-  },
+type TimingLookupArray = number[];
+interface TimingLookup {
+  [key: string]: TimingLookupArray;
+}
+
+const timingMapping: TimingFunctionMap = {};
+
+const timingLookup: TimingLookup = {
+  ease: [0.25, 0.1, 0.25, 1.0],
+  'ease-in': [0.42, 0, 1.0, 1.0],
+  'ease-out': [0, 0, 0.58, 1.0],
+  'ease-in-out': [0.42, 0, 0.58, 1.0],
+  'ease-in-sine': [0.12, 0, 0.39, 0],
+  'ease-out-sine': [0.12, 0, 0.39, 0],
+  'ease-in-out-sine': [0.37, 0, 0.63, 1],
+  'ease-in-cubic': [0.32, 0, 0.67, 0],
+  'ease-out-cubic': [0.33, 1, 0.68, 1],
+  'ease-in-out-cubic': [0.65, 0, 0.35, 1],
+  'ease-in-circ': [0.55, 0, 1, 0.45],
+  'ease-out-circ': [0, 0.55, 0.45, 1],
+  'ease-in-out-circ': [0.85, 0, 0.15, 1],
+  'ease-in-back': [0.36, 0, 0.66, -0.56],
+  'ease-out-back': [0.34, 1.56, 0.64, 1],
+  'ease-in-out-back': [0.68, -0.6, 0.32, 1.6],
 };
 
 const defaultTiming = (t: number): number => t;
 
 const parseCubicBezier = (str: string) => {
   //cubic-bezier(0.84, 0.52, 0.56, 0.6)
-  const regex = /cubic-bezier\((?:\s*(\d+\.?\d*)\s*,){3}\s*(\d+\.?\d*)\s*\)/;
+  const regex = /-?\d*\.?\d+/g;
   const match = str.match(regex);
 
   if (match) {
-    const [, num1, num2, num3, num4] = match.slice(1);
+    const [num1, num2, num3, num4] = match;
     const a = parseFloat(num1 || '0.42');
     const b = parseFloat(num2 || '0');
     const c = parseFloat(num3 || '1');
@@ -172,7 +170,7 @@ const parseCubicBezier = (str: string) => {
   }
 
   // parse failed, return linear
-  console.warn('Unknown timing function: ' + str);
+  console.warn('Unknown cubic-bezier timing: ' + str);
   return defaultTiming;
 };
 
@@ -187,10 +185,39 @@ export const getTimingFunction = (
     return timingMapping[str] || defaultTiming;
   }
 
+  if (str === 'linear') {
+    return (time: number) => {
+      return time;
+    };
+  }
+
+  if (str === 'step-start') {
+    return () => {
+      return 1;
+    };
+  }
+
+  if (str === 'step-end') {
+    return (time: number) => {
+      return time === 1 ? 1 : 0;
+    };
+  }
+
+  if (timingLookup[str] !== undefined) {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore - TS doesn't understand that we've checked for undefined
+    const [a, b, c, d] = timingLookup[str];
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+    const timing = getTimingBezier(a, b, c, d);
+    timingMapping[str] = timing;
+    return timing;
+  }
+
   if (str.startsWith('cubic-bezier')) {
     return parseCubicBezier(str);
   }
 
+  console.warn('Unknown timing function: ' + str);
   return defaultTiming;
 };
 
--- a/src/core/utils.ts
+++ b/src/core/utils.ts
@@ -23,8 +23,6 @@
  * @module
  */
 
-import memo from 'memize';
-
 export const EPSILON = 0.000001;
 export let ARRAY_TYPE =
   typeof Float32Array !== 'undefined' ? Float32Array : Array;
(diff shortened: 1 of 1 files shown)


frank-weindel (approved):
LGTM!

src/core/utils.ts
```diff
@@ -121,97 +119,80 @@ const getTimingBezier = (
...
-          return function (time) {
-            return time;
-          };
-        }
-    }
+interface TimingFunctionMap {
+  [key: string]: (time: number) => number | undefined;
+}
+
+const timingMapping: TimingFunctionMap = {
```
chiefcll, line 126: One minor tweak - I would create the timingMapping on the first call to getTimingFunction so this doesn't add overhead to the initial load of the site.

chiefcll:
Just one tweak - but nice research and optimizing 👍

chiefcll:
image

I've updated the test with the new code. Memo is still kicking but when there is one cubic. Let me know if you have the updated test code but I think we still want the memo in there.

chiefcll:
If you don't want the memo, you just need to cache the results in the timing map rather than returning it right away.

src/core/utils.ts
```diff
@@ -121,97 +119,80 @@ const getTimingBezier = (
...
+  if (str === '') {
+    return defaultTiming;
+  }
+
+  if (timingMapping[str] !== undefined) {
+    return timingMapping[str] || defaultTiming;
+  }
+
+  if (str.startsWith('cubic-bezier')) {
+    return parseCubicBezier(str);
```
chiefcll, line 191: rather than returning add the str to the timingMap and the result. Then next time it runs it will be part of the hash.

src/core/utils.ts
chiefcll: Actually - don't pre create this - add them to the timing map one at a time as they are used.

wouterlucas: I'd still have to store the bezier numbers that belong to a named instance, it will add an additional lookup which might not be at all that beneficial to the throughput. I don't think the overhead of creating that in an object at start is all that much, the bezier function is quite optimized and by the time you start doing a few calls I suspect it will be pretty efficient. But assumptions is the mother of all mistakes, so will give it a test.

src/core/utils.ts
wouterlucas: yeah thats exactly what's happening:
https://github.com/lightning-js/renderer/pull/249/files#diff-0be80b3dee46a398e160bb85c7da4446a1c5a3cac4a73aea3d908e00a24f2d2dR169

The if above that should then return the cached result.

or do you mean something else?

wouterlucas:
> image
> 
> I've updated the test with the new code. Memo is still kicking butt when there is one cubic. Let me know if you have the updated test code but I think we still want the memo in there.

I think you're running an older experiment, prior to caching being added. Feel free to pull the latest!

src/core/utils.ts
chiefcll: I mean the result of passing in a custom cubic-bezier should be cached in the timing map so it doesn't calculate again.

src/core/utils.ts
```diff
@@ -121,97 +119,80 @@ const getTimingBezier = (
...
+
+  if (match) {
+    const [, num1, num2, num3, num4] = match.slice(1);
+    const a = parseFloat(num1 || '0.42');
+    const b = parseFloat(num2 || '0');
+    const c = parseFloat(num3 || '1');
+    const d = parseFloat(num4 || '1');
+
+    const timing = getTimingBezier(a, b, c, d);
+    timingMapping[str] = timing;
```
chiefcll, line 169: Ah - I didn't see this. Cool so you are storing the result. Yup this works

src/core/utils.ts
chiefcll: I see you're doing it in the parse

chiefcll:
Just to summarize...
1. We're memoizing the cubic-beziers at startup with all the named ones, even though the user may not use any of them. Costing startup time.
2. You've added memoization for each custom cubic-bezier by storing it in the timingMap.

Ideally each cubic-bezier should be calculated and cached when they are needed (which is what the memoize library did). Doing it with the hash map is more efficient though (as its not a generic library doing it). https://www.npmjs.com/package/fast-memoize is supposed to be faster but I didn't have the time to switch that out when I added this. I think we should go back to the original code and just add the timingMap cache.

wouterlucas:
Going back to the original code will not get you close to the same results. It's even a little faster then memoization the entire function.

To verify this I've ran some tests including: hash map tests, lazy loading of the timings and adding the cache to the original function, here are the results:

3 named + 1 cubic:
```
[Log] original x 366 ops/sec ±6.87% (38 runs sampled) (index.ts, line 21)
[Log] original cached x 1,408 ops/sec ±1.61% (42 runs sampled) (index.ts, line 21)
[Log] memoized x 1,167 ops/sec ±1.74% (45 runs sampled) (index.ts, line 21)
[Log] optimized x 4,806 ops/sec ±0.76% (46 runs sampled) (index.ts, line 21)
[Log] optimized lazy x 4,504 ops/sec ±0.55% (47 runs sampled) (index.ts, line 21)
[Log] hashmap x 2,356 ops/sec ±1.13% (45 runs sampled) (index.ts, line 21)
[Log] hashmap lazy x 2,611 ops/sec ±0.37% (44 runs sampled) (index.ts, line 21)
```

And even though there is a small hit on lazy loading in total throughput, due to the additional IF checks required to make that happen. 

Fully agree that this might make loading/instantiation a lot faster.

For comparison a worst case run, 17 named + 30 cubics:
```
[Log] original x 256 ops/sec ±6.67% (39 runs sampled) (index.ts, line 21)
[Log] original cached x 1,509 ops/sec ±1.94% (44 runs sampled) (index.ts, line 21)
[Log] memoized x 237 ops/sec ±0.98% (45 runs sampled) (index.ts, line 21)
[Log] optimized x 3,021 ops/sec ±0.91% (47 runs sampled) (index.ts, line 21)
[Log] optimized lazy x 2,915 ops/sec ±0.71% (47 runs sampled) (index.ts, line 21)
[Log] hashmap x 2,258 ops/sec ±1.05% (45 runs sampled) (index.ts, line 21)
[Log] hashmap lazy x 2,308 ops/sec ±0.83% (47 runs sampled) (index.ts, line 21)
```

and only 2 named instances:
```
[Log] original x 2,701 ops/sec ±0.74% (46 runs sampled) (index.ts, line 21)
[Log] original cached x 2,497 ops/sec ±0.81% (47 runs sampled) (index.ts, line 21)
[Log] memoized x 1,413 ops/sec ±0.66% (46 runs sampled) (index.ts, line 21)
[Log] optimized x 4,819 ops/sec ±1.06% (47 runs sampled) (index.ts, line 21)
[Log] optimized lazy x 4,518 ops/sec ±0.61% (47 runs sampled) (index.ts, line 21)
[Log] hashmap x 2,519 ops/sec ±0.85% (47 runs sampled) (index.ts, line 21)
[Log] hashmap lazy x 2,848 ops/sec ±1.50% (45 runs sampled) (index.ts, line 21)
```

To summarize:
* No need for an external library, implementing caching where it is needed by ourselves is faster
* Writing optimized JIT-friendly code is fastest
* Lazy loading has a small hit, but can help on load times

Other then that hashmaps seem to be slower then object key/values caching on the box, kind of surprised me to.
Tests are pushed to the experiments repo.

If desired I can push the updated PR with the lazy loading implementation.

wouterlucas:
Re the hashmaps, definitely something we might want to track. I suspect this might be due to certain optimization layers not being available in 32-bit on WPEWebKit 2.38. If I run the same tests in 64-bit I get a lot different results, there the hashmaps start to really take off and even the lazy loading going 🚀 

Something worthwhile to discuss with the webkit friends, and we should test this again once the new WebKit is made available to us.

frank-weindel (approved):
Is creating new functions each time for `linear`, `step-start`, and `step-end` really faster than re-using one copy of the function from a map cache? I'll move this forward cause this has gone on way too long

wouterlucas:
> Is creating new functions each time for `linear`, `step-start`, and `step-end` really faster than re-using one copy of the function from a map cache? I'll move this forward cause this has gone on way too long

It is not faster no, but then we go back to the "initial hit" on startup concern that was raised by Chris. This is a sort of middle ground to balance performance vs initial creation hit.

Funny enough it does get faster to lazy load when FTL becomes available in 32-bit. So if you run the experiment on a 64 bit browser you'll see the "lazy" loading approaches get faster. Same as hashmaps, though marginally compared to what we do now.

We could revisit this once those optimizations in the 32-bit browser versions are available to the RDK community.

PR #252, opened 2024-04-17
Simplify viewport events

wouterlucas:
A node in the visible area gets a inViewport event A node in bounds gets a inBounds event
A node outside viewport & bounds gets a outOfViewport event

Nodes that are init or outofview port will get a renderable false. All events have a previous state and new state accordingly.

**BREAKING CHANGE**

### DIFF
--- a/src/core/CoreNode.ts
+++ b/src/core/CoreNode.ts
@@ -17,11 +17,7 @@
  * limitations under the License.
  */
 
-import {
-  assertTruthy,
-  mergeColorAlphaPremultiplied,
-  getImageAspectRatio,
-} from '../utils.js';
+import { assertTruthy, mergeColorAlphaPremultiplied } from '../utils.js';
 import type { ShaderMap } from './CoreShaderManager.js';
 import type {
   ExtractProps,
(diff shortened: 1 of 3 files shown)


src/core/CoreNode.ts
```diff
@@ -57,14 +53,14 @@ import { RenderCoords } from './lib/RenderCoords.js';
 
 export enum CoreNodeRenderState {
   Init = 0,
-  OutOfBounds = 2,
+  OutOfViewport = 2,
```
frank-weindel, line 56: @wouterlucas I think this needs to remain OutOfBounds. OutOfViewport implies that anywhere outside of the viewport which could even be the Bounds Margin. Consider the diagram below:
Screen Shot 2024-04-17 at 11 46 38 AM

wouterlucas: That works too :) I'll flip it back to Out of Bounds. My mind went the other route, making bounds an intermediate step and viewport something that an element is either in or out of.

jfboeve (approved):

frank-weindel (approved):

PR #253, opened 2024-04-22
Fix Canvas2D fonts performance

wouterlucas:
This PR adds a "single page" Canvas2D text rendering path for small Canvas text usage along side of the original "scrollable" Canvas2D text path.


src/core/text-rendering/renderers/CanvasTextRenderer.ts
```diff
@@ -714,8 +814,23 @@ export class CanvasTextRenderer extends TextRenderer<CanvasTextRendererState> {
   override destroyState(state: CanvasTextRendererState): void {
     // Remove state object owner from any canvas page textures
     state.canvasPages?.forEach((pageInfo) => {
-      pageInfo.texture?.setRenderableOwner(state, false);
+      const { texture } = pageInfo;
```
frank-weindel, line 817: Something important to note here is that `destroyState()` is only called when a Text Node is destroyed or a new text renderer is loaded by a Text Node via the `textRendererOverride` property. If new text is set (or any property that triggers a new Canvas draw are set) on a single persistent Text Node, the old texture(s) will not be freed from what I've gathered.

src/core/text-rendering/renderers/CanvasTextRenderer.ts
```diff
@@ -285,12 +373,15 @@ export class CanvasTextRenderer extends TextRenderer<CanvasTextRendererState> {
...
       renderInfo: undefined,
+      canvasPage: undefined,
+      canvasPages: undefined,
       forceFullLayoutCalc: false,
       textW: 0,
       textH: 0,
       fontInfo: undefined,
       fontFaceLoadedHandler: undefined,
       isRenderable: false,
+      isScrollable: props.scrollable === true,
```
frank-weindel, line 384: Is it necessary to make this a state property? If a user were to change the `scrollable` property of a Text Node mid-lifecycle it would not have any affect.

src/core/CoreTextureManager.ts
```diff
@@ -266,30 +292,40 @@ export class CoreTextureManager {
...
+
+  /**
+   * Remove a `Texture` from the texture cache by its cache key
+   *
+   * @remarks
+   * Same as removeTextureIdFromCache, but then by cache key.
+   *
+   * @param cacheKey
+   */
+  removeTextureCacheKeyFromCache(cacheKey: string): void {
```
frank-weindel, line 316: I think there's a misunderstanding (very understandably) regarding the texture caches. They primarily exist to efficiently map a TextureRef from the Main space with a Texture on the Core space. A concept required to support ThreadX. Every TextureRef gets a unique ID assigned to it, which is the id number you see mentioned in the `textureIdCache`.

Because the CanvasTextRenderer operates 100% within the Core space, it doesn't work with TextureRefs and hence doesn't work with these IDs. Hence when it calls `loadTexture()`, the Texture that is returned from it isn't cached at all in those caches. Notice this in `loadTexture()`:
```
      if (descId) {
        this.addTextureIdToCache(descId, cacheKey, texture);
      }
```

So I don't think you need to worry at all about removing these textures from the texture cache. You really just want to make sure `free()` is called on the CoreContextTexture associated with the Texture (it's a 1-1 mapping) so that it frees the GPU memory right away. The Texture/CoreContextTexture instances themselves, being no longer referenced, should be released in the normal GC flow.

wouterlucas: got it! I started down this path and later indeed figured I could just grab the context directly and call free on it. I'll clean it up.

src/core/text-rendering/renderers/CanvasTextRenderer.ts
wouterlucas: that's exactly why. I figured it might be better to stay on the 3 canvasPage rendering instead of going back to the single, to avoid large discrepancies with rendering and positioning. If you feel that would be a better path, easy to flip it back to just `state.props.scrollable` and not have this here.

that said would this really be a usecase? a user repurposing a large chunk of scrollable text to a single non-scrollable piece of text?

src/core/text-rendering/renderers/CanvasTextRenderer.ts
wouterlucas: mm yes good point - I was looking into invalidating caches today with a bug that wouldn't apply a color after Text Node creation. I guess once we clear the canvasPage caches because we're going to redraw we should also clear out the textures, thanks!

src/core/textures/ImageTexture.ts
```diff
@@ -160,6 +160,18 @@ export class ImageTexture extends Texture {
     return `ImageTexture,${key},${resolvedProps.premultiplyAlpha ?? 'true'}`;
   }
 
+  override free(): void {
+    if (this.props.src instanceof ImageData) {
+      // ImageData is a non-cacheable texture, so we need to free it manually
```
elsassph, line 165: This is incorrect since the addition of the image factory/key PR #256

frank-weindel:
Replaced by #319

PR #257, opened 2024-04-23
Text Rendering: Reported Text Height Changes (Phase 1+2+3)

### LINKED ISSUE #157: Canvas Text: Reported dimensions do not include descenders

### LINKED ISSUE #170: Text Rendering: Reported Text Height Change Proposal

frank-weindel:
- Make reported Text Height dimension of text based purely on Line Height and number of lines rendered (plus offsetY)
- Calculate the default lineHeight from font metrics
  - Font metrics can now be provided in font definitions for accurate line height and baseline offsets.
- Make the default verticalAlign text property value "middle"

BREAKING CHANGE
- Constructors for WebTrFontFace and SdfTrFontFace have changed to use objects and add the ability to provide font metrics which are REQUIRED for SDF fonts and HIGHLY RECOMMENDED for Web/Canvas fonts. Use the L3 [msdf-generator](https://github.com/lightning-js/msdf-generator/) to get these metrics.
- Text alignment will be different because of these metrics and the now defaulting of`verticalAlign` to 'middle'. Please try to adjust to this new behavior instead of trying to change them back to how they were.

Fixes #157
Fixes #170


frank-weindel:
There is a very subtle bug that is preventing me from producing the correct version of the `text-vertical-layout-1` test. Something is causing certain nodes to not be rendered at consistent times in mysterious ways.

frank-weindel:
Depends on:
- #263
- #261

jfboeve (approved):

wouterlucas (approved):
Nice! @michielvandergeest / @uguraslan you guys should align `msdfGenerator.js` with https://github.com/lightning-js/msdf-generator and/or make sure the font metrics are always included.

Judging by what Frank provided w.r.t. the browser support for [TextMetrics](https://developer.mozilla.org/en-US/docs/Web/API/TextMetrics) the support isn't going to be great. Having those pregenerated will be an important feature.

PR #267, opened 2024-05-02
Optimize Matrix3d storage and operations

frank-weindel:
Matrix3d now stores its values properties instead of a Float32Array. This should potentially cut down on any performance penalty between transforming native JavaScript numbers to Float32 values and vice-versa.

The bottom row of the matrix is now considered a constant [0, 0, 1] which will be true for all 2D operations. The bottom row values hence no longer are stored in the Matrix3d instance.

Also because of the bottom-row assumption the `multiply` operation no longer needs to calculate the bottom row.

The Float32Array, needed by SDF Text Nodes is now only generated (and updated) when needed. This should significantly reduce the number of arrays allocated in an application.

- [x] Pending performance and memory review

### DIFF
--- a/src/core/lib/Matrix3d.ts
+++ b/src/core/lib/Matrix3d.ts
@@ -1,71 +1,69 @@
-/* eslint-disable @typescript-eslint/no-non-null-assertion */
-
-// Matrix3d is a 3x3 matrix in column-major order because that's how WebGL likes it.
-// The matrix is stored in a Float32Array in the following order:
-// | 0 3 6 |
-// | 1 4 7 |
-// | 2 5 8 |
-// The following constants are used to index into the array in a row-major way.
-const m0 = 0;
-const m1 = 3;
-const m2 = 6;
-const m3 = 1;
-const m4 = 4;
-const m5 = 7;
-const m6 = 2;
-const m7 = 5;
-const m8 = 8;
+/*
+ * If not stated otherwise in this file or this component's LICENSE file the
+ * following copyright and licenses apply:
+ *
+ * Copyright 2023 Comcast Cable Communications Management, LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the License);
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


frank-weindel:
# Memory Analysis

The new Matrix3d offered in this PR saves a significant amount of memory per Node depending on if the Node is an SDF Node or any other type of Node.

| Node Type                        | Memory % of Old | Savings % |
| -------------------------------- | --------------- | --------- |
| Best Case (Per Non-SDF Node)     | 40%             | **60%**   |
| Worst Case (Per SDF Node)        | 71%             | **29%**   |

Memory snapshots taken on Chrome v124 from the `no-rtt` test example (No SDF Text Used)

## Non-SDF Nodes

|        | Type + Quantity                    | Retained Size   |
| -------| ---------------------------------- | --------------- |
| Before | Matrix3d x6013                     | 1,178,548 bytes |
| After  | Matrix3d (w/o Float32Array) x6013  | 480,956 bytes   |

The new Matrix3d (w/o the conditionally generated Float32Array for SDF text) retains **40%** the memory that the old one did.

## With/Without Float32Array

The new Matrix3d class will now only create a Float32Array when one is needed:

|        | Type + Quantity                    | Retained Size   |
| ------ | ---------------------------------- | --------------- |
| Before | Matrix3d x1                        | 196 bytes       |
| After  | Matrix3d x1 (w/o Float32Array)     | 80 bytes        |
|        | Matrix3d x1 (w/ Float32Array)      | 260 bytes       |

## SDF Nodes

With the Float32Array, the new Matrix3d class is bigger than the original. However, there are 3 matrices created/managed by each Node. Only 1 of the 3 (in the case of SDF Text Nodes) will ever include the Float32Array:

|        | Type                               | Retained Size   |
| ------ | ---------------------------------- | --------------- |
| Before | Matrix3d (per SDF Node)            | 588 bytes       |
| After  | Matrix3d (per SDF Node)            | 420 bytes       |

So in the worst case where all the Nodes rendered are SDF Text Nodes, the new Matrix3d retains **71%** of the memory that the old one did.

elsassph:
Nice! Would there be a benefit in doing a similar change to `RenderCoords`?

frank-weindel:
> Nice! Would there be a benefit in doing a similar change to `RenderCoords`? BTW it looks like `calculateRenderCoords` and `updateBoundingRect` could potentially be combined, for efficiency.

Very possibly. @jfboeve, is this something you could look into perhaps? 🙏

erikhaandrikman (approved):

PR #270, opened 2024-05-07
feat: children supporting order

chiefcll:
Update renderer for children to support order of nodes.

### DIFF
--- a/src/render-drivers/main/MainOnlyNode.ts
+++ b/src/render-drivers/main/MainOnlyNode.ts
@@ -368,8 +368,6 @@ export class MainOnlyNode extends EventEmitter implements INode {
 
   set parent(newParent: MainOnlyNode | null) {
     const oldParent = this._parent;
-    this._parent = newParent;
-    this.coreNode.parent = newParent?.coreNode ?? null;
     if (oldParent) {
       const index = oldParent.children.indexOf(this);
       assertTruthy(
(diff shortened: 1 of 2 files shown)


src/render-drivers/threadx/ThreadXMainNode.ts
```diff
@@ -165,7 +164,21 @@ export class ThreadXMainNode extends SharedNode implements INode {
...
+    node.parent = this;
+    node.parentId = this?.id ?? 0;
+
+    if (beforeNode) {
+      const index = this.children.indexOf(beforeNode);
+      assertTruthy(
+        index !== -1,
+        'ThreadXMainNode.beforeNode: Node not found in children!',
+      );
+      this.children.splice(index, 0, node);
```
frank-weindel, line 181: The ThreadX side of this needs some care. To support this, we'd likely need to use messages to communicate parent changes. But then we're getting into talking about having all mutations be communicated my messages in order to ensure everything happens in order 🤔

src/render-drivers/main/MainOnlyNode.ts
```diff
@@ -379,14 +377,28 @@ export class MainOnlyNode extends EventEmitter implements INode {
...
+    node.coreNode.parent = this?.coreNode ?? null;
+
+    if (beforeNode) {
+      const index = this.children.indexOf(beforeNode);
+      assertTruthy(
+        index !== -1,
+        'MainOnlyNode.beforeNode: Node not found in children!',
+      );
+      this.children.splice(index, 0, node);
+    }
```
elsassph, line 399: Is it missing `else push`?

elsassph: Does it?

chiefcll:
Since children are sorted by zIndex this won't work... will need a separate array

PR #271, opened 2024-05-09
Feat: Add distort shader

lewispeel:
I recently created a shader In Lightning 2 to distort a texture inside itself and ported it to Lightning 3. It allows you to define new coordinates for top-left, top-right, bottom-right and bottom-left inside it's original dimensions.

## Changes
- Added `DistortShader` to WebGL shaders (based on https://www.shadertoy.com/view/lsVSWW).
- Added `Point` interface to `CommonTypes` to make shader params typesafe.
- Added an example in `examples/tests/distort-shader` that shows using pixel values and normalized values.
- Updated `CoreShaderManager` to include `DistortShader` as an option.

## How to test
- Run `pnpm start`
- In the browser, navigate to `http://localhost:5173/?test=distort-shader`

## Usage
You can use explicit pixel values or normalized values e.g. `{ x: 0.5 }` will be 50% of the texture width. To use normalized values, add `normalized: true` to the shader parameters.

```js
const size = 500;

const parent = renderer.createNode({
  parent: testRoot,
});

// distorted using pixel values
const pixels = renderer.createNode({
  x: 100,
  y: 100,
  width: size,
  height: size,
  parent,
  src: `https://picsum.photos/${size}/${size}`,
  shader: renderer.createShader('DistortShader', {
    topLeft: { x: 20, y: 50 },
    topRight: { x: 400, y: 0 },
    bottomRight: { x: 350, y: 300 },
    bottomLeft: { x: 50, y: 350 },
  }),
});

// distorted using normalized values
const normalized = renderer.createNode({
  x: 700,
  y: 100,
  width: size,
  height: size,
  parent,
  src: `https://picsum.photos/${size}/${size}`,
  shader: renderer.createShader('DistortShader', {
    normalized: true,
    topLeft: { x: 0.1, y: 0.1 },
    topRight: { x: 0.75, y: 0.2 },
    bottomRight: { x: 0.95, y: 0.7 },
    bottomLeft: { x: 0.05, y: 0.9 },
  }),
});
```

## Output
Examples using pixel values (left), normalized values (right). The dark rectangle are to illustrate where the original dimensions of the image textures.

image


src/core/renderers/webgl/shaders/Distort.ts
```diff
@@ -0,0 +1,227 @@
...
+    `,
+    fragment: `
+    # ifdef GL_FRAGMENT_PRESICISON_HIGH
+    precision highp float;
+    # else
+    precision mediump float;
+    # endif
+
+    uniform sampler2D u_texture;
+    uniform vec2 u_dimensions;
```
frank-weindel, line 166: Could you normalize the formatting/indentation between these two shaders? 2 spaces per indent. First line of GLSL code indented 1 level past the indentation of the previous line of JavaScript 🙏

src/core/renderers/webgl/shaders/Distort.ts
```diff
@@ -0,0 +1,227 @@
...
+  bottomRight?: Point;
+
+  /**
+   * x & y coordinates of the bottom left point
+   * @default { x: 0, y: 1080 }
+   */
+  bottomLeft?: Point;
+}
+
+export class Distort extends WebGlCoreShader {
```
frank-weindel, line 60: For the sake of uniformity and clarity going forward, could you rename this (as well as adjusting the related types/strings) to `DistortShader`? The name for `RoundedRectangle` is an artifact which we should probably update.

src/render-drivers/utils.ts
```diff
@@ -58,7 +58,12 @@ export async function loadCoreExtension(
 }
 
 export function santizeCustomDataMap(d: CustomDataMap): CustomDataMap {
-  const validTypes = { boolean: true, string: true, number: true, undefined: true };
+  const validTypes = {
```
frank-weindel, line 61: Can you back these lines out in a commit amend/force push? That would be great as to not disrupt the original line attribution.

src/core/renderers/webgl/shaders/Distort.ts
```diff
@@ -0,0 +1,227 @@
...
+  }
+
+  static z$__type__Props: DistortProps;
+
+  static override resolveDefaults(props: DistortProps): Required<DistortProps> {
+    return {
+      topLeft: props.topLeft || { x: 0, y: 0 },
+      topRight: props.topRight || { x: 1920, y: 0 },
+      bottomRight: props.bottomRight || { x: 1920, y: 1080 },
+      bottomLeft: props.bottomLeft || { x: 0, y: 1080 },
```
frank-weindel, line 85: These defaults are a bit too specific. I think one thing you might be able to do to achieve this dynamically is to allow `null` as value for each of these keys and make `null` the default. Then in `bindProps` I think you could check for `null` and dynamically create the defaults using the `$dimensions` property which at that point is reliable. (Going to leave a separate comment in regards to your request for comments on potentially using normalized 0.0 -> 1.0 values)

frank-weindel:
@lewispeel Awesome work. Thanks for the contribution. Regarding your question about potentially using relative normalized values instead of absolute position values, I'd say allow both. I can see both methods coming in handy depending on the situation. You could base it off of a new boolean prop flag (`normalized` perhaps) and do all the necessary resolutions in `bindProps()`.

Please let me know if you need assistance generating the certified snapshot needed for this. You need to run the visual regression test runner in `--ci --capture` mode.

src/core/renderers/webgl/shaders/Distort.ts
lewispeel: Done.

examples/tests/distort.ts
```diff
@@ -0,0 +1,41 @@
+import type { ExampleSettings } from '../common/ExampleSettings.js';
+
+export async function automation(settings: ExampleSettings) {
+  // Snapshot single page
+  await test(settings);
+  await settings.snapshot();
+}
+
+export default async function test({ renderer, testRoot }: ExampleSettings) {
```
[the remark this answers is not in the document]
lewispeel: Done.

src/core/renderers/webgl/shaders/Distort.ts
lewispeel: Done. To get this working using `null` I had to modify the `Point` class to accept number and null types. Allowing `null` feels a bit weird to me. I could keep it as number only and the default be zero, then assume if I see a zero size width or height value assume its been left as zero on purpose and use the full width/height? What do you think?

frank-weindel: I'd regard the whole Point as a unit. The type of each corner in that cause would be `Point | null`. So if you need to provide an `x`, you also need to provide the `y`.

lewispeel: Ah that makes sense, I'll change it 👍

lewispeel: @frank-weindel Done!

src/core/renderers/webgl/shaders/DistortShader.ts
```diff
@@ -0,0 +1,246 @@
...
+    const bottomLeft = [
+      (props.bottomLeft.x || 0) / width,
+      (props.bottomLeft.y || height) / height,
+    ];
+
+    // strange bug
+    // if the dimensions create a rectangle, nothing is rendered
+    // adding a small amount to one of the x values resolves this
+    if (topLeft[0] === bottomLeft[0] && topRight[0] === bottomRight[0]) {
+      topLeft[0] = +0.0001;
```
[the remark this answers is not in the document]
lewispeel: Good catch, thanks @erikhaandrikman - this has been updated.

src/core/renderers/webgl/shaders/DistortShader.ts
```diff
@@ -0,0 +1,246 @@
...
+      uniform vec2 u_bottomRight;
+
+      varying vec2 v_textureCoordinate;
+      varying vec4 v_color;
+
+      float xross(in vec2 a, in vec2 b) {
+        return a.x * b.y - a.y * b.x;
+      }
+
+      vec2 invBilinear(in vec2 p, in vec2 a, in vec2 b, in vec2 c, in vec2 d ){
```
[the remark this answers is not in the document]
lewispeel: Amazing! Thanks @erikhaandrikman - this is fixed.

frank-weindel:
@lewispeel This needs a snapshot generated with:
```
pnpm test:visual --ci --capture
```

Check out the `DOCKER.md` file for more information on how to set things up.

lewispeel:
@erikhaandrikman I've just spotted a problem.. this demo works fine on single nodes with no children, but if I create a node with child nodes (requiring `rtt=true`), the texture is flipped vertically... `topLeft` becomes `bottomLeft` etc. Do you know any reason why this might be?

lewispeel:
> @lewispeel This needs a snapshot generated with:
> 
> ```
> pnpm test:visual --ci --capture
> ```
> 
> Check out the `DOCKER.md` file for more information on how to set things up for it.

Thanks Frank, I'll wait for Erik to feed back on the issue above and I'll make sure to run the visual tests 👍

erikhaandrikman:
> @erikhaandrikman I've just spotted a problem.. this demo works fine on single nodes with no children, but if I create a node with child nodes (requiring `rtt=true`), the texture coords are flipped vertically... `topLeft` becomes `bottomLeft` `topRight` becomes `bottomRight`. Do you know any reason why this might be?
> 
> To test this, comment out lines 93-95 in `tests/distort-shader.ts` - it should apply the effect to only the top left but applies it to the bottom left. If you then comment out `rtt:true` on line 89 and 99-133 and the effect works again.

probably due to the fact that the render texture is render flipped ( and coords being before pushing them to the buffer: https://github.com/lightning-js/renderer/blob/70af9d42091c1edfd9325adb9ba7399945fe6280/src/core/renderers/webgl/WebGlCoreRenderer.ts#L306 )

i can look into this

erikhaandrikman:
We're encountering some projection issues, primarily with text, when attempting to flip the rendering in render texture nodes. As a temporary solution, we've been rendering the contents to an extra childNode (with RTT enabled) and flipping the rendering on the Y-axis by setting `scaleY: -1`

```
const distorted = renderer.createNode({
    x: 700,
    y: 100,
    width: size,
    height: size,
    parent,
    src: `https://picsum.photos/${size}/${size}`,
    shader: renderer.createShader('DistortShader', {
      normalized: true,
      topLeft: { x: 0.1, y: 0.1 },
      topRight: { x: 0.75, y: 0.2 },
      bottomRight: { x: 0.95, y: 0.7 },
      bottomLeft: { x: 0.05, y: 0.9 },
    }),
  });

  // distorted children inside a distorted container (requires rtt=true)
  const nested = renderer.createNode({
    x: 1300,
    y: 100,
    width: size,
    height: size,
    parent,
    rtt: true,
  });

  const nestedInner = renderer.createNode({
    width: size,
    height: size,
    parent: nested,
    rtt: true,
    color:0xffffffff,
    shader: renderer.createShader('DistortShader', {
      normalized: true,
      topLeft: { x: 0.1, y: 0.1 },
      topRight: { x: 0.75, y: 0.2 },
      bottomRight: { x: 0.95, y: 0.7 },
      bottomLeft: { x: 0.05, y: 0.9 },
    }),
    scaleY: -1,
  });

  const nestedBg = renderer.createNode({
    width: size,
    height: size,
    color: 0x990000ff,
    parent: nestedInner
  });

  const nestedHolder = renderer.createNode({
    width: size,
    height: size,
    parent: nestedInner,
    scaleY:-1,
    color: 0x99000000
  });
  
  // a distorted image inside a distorted container
  const image = renderer.createNode({
    x: size / 2,
    y: size / 2,
    width: size / 2,
    height: size / 2,
    mount: 0.5,
    parent: nestedHolder,
    src: `https://picsum.photos/${size}/${size}`,
    shader: renderer.createShader('DistortShader', {
      normalized: true,
      topLeft: { x: 0.1, y: 0.1 },
      topRight: { x: 0.75, y: 0.2 },
      bottomRight: { x: 0.95, y: 0.7 },
      bottomLeft: { x: 0.05, y: 0.9 },
    }),
  });
```

wouterlucas:
@erikhaandrikman can we move this along with known limitations of RTT?

PR #273, opened 2024-05-12
Add ability to name VRT snapshots

frank-weindel:
Calls to snapshot() now accept a `name` key in the options object. This name, if provided, is appended to the end of the test name and used in the snapshot file name.


wouterlucas (approved):
LGTM

PR #274, opened 2024-05-12
Animation 'animating' and 'stopped' events

frank-weindel:
- IAnimationController is now an event emitter.
- `waitUntilStarted()` has been replaced by the 'animating' event
- Added 'stopped' event.
- `waitUntilStopped()` promised is now created more consistently
- Excess `dt` that depletes a delay is now applied to the animation progress


src/core/animations/CoreAnimation.ts
```diff
@@ -96,24 +96,40 @@ export class CoreAnimation extends EventEmitter {
...
-  applyEasing(p: number, s: number, e: number): number {
+  private applyEasing(p: number, s: number, e: number): number {
     return (this.timingFunction(p) || p) * (e - s) + s;
   }
 
   update(dt: number) {
     const { duration, loop, easing, stopMethod } = this.settings;
-    if (!duration) {
+    const { delayFor } = this;
+    if (!duration && !delayFor) {
```
wouterlucas, line 106: You may want to consider checking for `delayFor === 0` and `duration <= 0` as they are slightly cheaper then a complete falsy check. A falsy checks for all types, which just takes a few cycles more. Since this is a heavy traversed section of the code, all the little things help.

src/core/animations/CoreAnimation.ts
```diff
@@ -96,24 +96,40 @@ export class CoreAnimation extends EventEmitter {
...
+        return;
+      } else {
+        // We went beyond the delay time, add it back to dt so we can continue
+        // with the animation.
+        dt = -this.delayFor;
+        this.delayFor = 0;
+      }
+    }
+
+    if (!duration) {
```
wouterlucas, line 124: same here

src/core/animations/CoreAnimation.ts
```diff
@@ -96,24 +96,40 @@ export class CoreAnimation extends EventEmitter {
...
 
     if (this.delayFor > 0) {
       this.delayFor -= dt;
+      if (this.delayFor >= 0) {
+        // Either no or more delay left. Exit.
+        return;
+      } else {
+        // We went beyond the delay time, add it back to dt so we can continue
+        // with the animation.
+        dt = -this.delayFor;
```
wouterlucas, line 119: 😱 nice catch!

src/core/animations/CoreAnimationController.ts
```diff
@@ -24,58 +25,56 @@ import type {
...
-    this.animation.once('start', this.started.bind(this));
-
-    this.makeStoppedPromise();
-    this.animation.once('finished', this.finished.bind(this));
-
-    // prevent registering the same animation twice
-    if (!this.manager.activeAnimations.has(this.animation)) {
-      this.manager.registerAnimation(this.animation);
+    if (this.state !== 'running') {
+      this.makeStoppedPromise();
```
wouterlucas, line 57: do we still need this? feels like there are two ways to achieve the same thing now, creating a promise and registering for events.

wouterlucas (approved):
Nice!

src/core/animations/CoreAnimationController.ts
frank-weindel: Yeah similar to the Web Animations API https://developer.mozilla.org/en-US/docs/Web/API/Animation/finished

It is still a useful way to convey a chain of animations, which would be harder to achieve with the event approach.

frank-weindel:
@wouterlucas Just removed the falsy checks. Had to refactor how the `settings` object is stored so that now it's now monomorphic and has all undefined values resolved to defaults. That itself could help a tiny bit on performance 🤞

PR #276, opened 2024-05-17
Fix assertion error thrown in WebGlCoreCtxTexture getter

### LINKED ISSUE #275: Assertion being thrown in WebGlCoreCtxTexture get ctxTexture.
kemin-issue

Reported by @pecoram

frank-weindel:
The issue occurs when a texture is loading, free() is called on it by the GC, and then there is an error thrown by getTextureData() from the Texture source.

Fixes #275


frank-weindel:
@pecoram to confirm if this actually fixes the issue he reported.

pecoram (approved):

pecoram:
tested.. it works

PR #277, opened 2024-05-20
Fix: missing CommonJS export

### LINKED ISSUE #258: Typescript `node` module resolution
### Issue

When `moduleResolution` is set to `node` (it can be quite a breaking change to update), the `main` entry from `package.json` will be used. Unfortunately it currently refers to a non existent file:

```jsonc
// package.json
{
  //...
  "main": "./dist/exports/index.js"
  //...
}
```

### Solution

Adding missing `exports/index.ts` seems to work (at least with a bundler):
```
export * from './main-api.js';
export * from './core-api.js';
export * from './utils.js';
```

Chances are though that something depending on CommonJS would fail (like Jest).

elsassph:
Fixes #258 - this creates the missing cjs export point as declared already in `package.json`:
```
"main": "./dist/exports/index.js",
```

### DIFF
--- a/exports/index.ts
+++ b/exports/index.ts
@@ -1,0 +1,22 @@
+/*
+ * If not stated otherwise in this file or this component's LICENSE file the
+ * following copyright and licenses apply:
+ *
+ * Copyright 2023 Comcast Cable Communications Management, LLC.
+ *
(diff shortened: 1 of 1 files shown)


wouterlucas (approved):
👍

PR #279, opened 2024-05-24
fix: numImageWorkers=0

pecoram:
When numImageWorkers is 0 the imageWorkerManager was still initizlizated


wouterlucas (approved):
Thanks! 👍 

Out of curiosity, what are cases when you'd like to turn off the image workers even though its available?

PR #280, opened 2024-05-27
Remove internal Drivers and ThreadX support

wouterlucas:
ThreadX is a feature implemented to support offscreen canvas multithreading, however browser support for ThreadX will not come soon to our space (TV / STB market) and the current implementation has some negative side effects.

With the introduction of Drivers there is a lot of double housekeeping, maintaining two tree's of nodes, the `DriverNode` and `CoreNode`. This causes additional processing overhead (mainly extra function invocations and emitters) and an additional list of nodes stored in memory.

In the future we should ensure all ThreadX specifics are retained in the ThreadX module, likely by providing a RendererMain replacement from the ThreadX project and proxy the CoreNode respectively as we communicate over threads. And keep the specifics to ThreadX to where they belong.

Early test results:
current master: 20k nodes create 2867ms @ 65 MB HEAP:
image_2024_05_27T19_51_20_477Z image_2024_05_27T19_52_05_633Z

this pr: 20k nodes create 1272ms @ 35 MB HEAP: 
image_2024_05_27T19_51_27_553Z image_2024_05_27T19_52_11_967Z

About 2.2x faster and nearly 50% less memory on an excessive stress test.

Todo:
- [x] Fix RTT
- [ ] Fix/review VRT test results
- [ ] Run an app on it


pecoram:
#281

erikhaandrikman:
🎉

wouterlucas:
Latest results with the PR compared to latest release:
image

PR #288, opened 2024-06-04
Fix image src conversion with local file paths (#287)

### LINKED ISSUE #287: Support for file:// in images
Certain TVs require apps to be loaded through `file://` and this causes an issue with the `location.href` not being usable for relative to absolute path detection. Causing images to not load properly with the following error: `<console.log> Error: Failed to load image:`

wouterlucas:
Support for `file://` loads with relative to absolute path conversion. 
Fixes #287

### DIFF
--- a/src/core/lib/utils.ts
+++ b/src/core/lib/utils.ts
@@ -248,3 +248,28 @@ export function isBoundPositive(bound: Bound): boolean {
 export function isRectPositive(rect: Rect): boolean {
   return rect.width > 0 && rect.height > 0;
 }
+
+export function convertUrlToAbsolute(url: string): string {
+  // handle local file imports
+  if (self.location.protocol === 'file:') {
+    const path = self.location.pathname.split('/');
+    path.pop();
+    const basePath = path.join('/');
+    const baseUrl = self.location.protocol + '//' + basePath;
+
+    // check if url has a leading dot
+    if (url.charAt(0) === '.') {
+      url = url.slice(1);
+    }
+
+    // check if url has a leading slash
+    if (url.charAt(0) === '/') {
+      url = url.slice(1);
+    }
+
+    return baseUrl + '/' + url;
+  }
+
+  const absoluteUrl = new URL(url, self.location.href);
+  return absoluteUrl.href;
+}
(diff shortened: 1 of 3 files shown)


src/core/lib/utils.ts
```diff
@@ -248,3 +248,28 @@ export function isBoundPositive(bound: Bound): boolean {
...
+
+    // check if url has a leading slash
+    if (url.charAt(0) === '/') {
+      url = url.slice(1);
+    }
+
+    return baseUrl + '/' + url;
+  }
+
+  const absoluteUrl = new URL(url, self.location.href);
```
frank-weindel, line 273: Is the main issue here that those TV's that only support `file://` have a problematic implementation of `URL`? I tested this in a `file://` loaded page in Chrome and this `new URL()` statement works whether the protocol is `file:` or `http:`.

src/core/lib/utils.ts
```diff
@@ -248,3 +248,28 @@ export function isBoundPositive(bound: Bound): boolean {
...
+    const baseUrl = self.location.protocol + '//' + basePath;
+
+    // check if url has a leading dot
+    if (url.charAt(0) === '.') {
+      url = url.slice(1);
+    }
+
+    // check if url has a leading slash
+    if (url.charAt(0) === '/') {
+      url = url.slice(1);
```
frank-weindel, line 267: Unlike `new URL()` this will not handle `..` appropriately

src/core/lib/utils.ts
wouterlucas: The URL works fine, it's the `location.href` that doesn't work when you load it through `file://`. As such we have to construct the base URL manually from the entire file path just like L2 did, using `URL` will set it to the highest top level directory on the filepath, something like `\\assets\rocko.png` while you're loading it from `\\c\Users\blabla\workspace\renderer\` or similar.

Filepaths != URLs as you do not have a host and port and only a path.

src/core/lib/utils.ts
wouterlucas: Fair - this will need to be manually checked and popped then, something L2 didn't do as well. I'd rather move this along and add the `..` handling later.

frank-weindel (approved):

PR #294, opened 2024-06-06
Remove INode

frank-weindel:


wouterlucas:
I guess Node is a reserved word? If so please stick to CoreNode, no need to change it.

frank-weindel:
> I guess Node is a reserved word? If so please stick to CoreNode, no need to change it.

Reverted

wouterlucas (approved):
Thanks!

PR #295, opened 2024-06-06
passing renderCoords from CoreNode to addQuad

jfboeve:
We are using the same calculations for the renderCoords we use in the .addQuad function in the WebGLCoreRenderer, with this fix we can use the already calculated renderCoords without having to do the calculations again every time a draw call is made.

### DIFF
--- a/src/core/renderers/webgl/WebGlCoreRenderer.ts
+++ b/src/core/renderers/webgl/WebGlCoreRenderer.ts
@@ -234,6 +234,7 @@ export class WebGlCoreRenderer extends CoreRenderer {
       tb,
       tc,
       td,
+      renderCoords,
       rtt: renderToTexture,
       parentHasRenderTexture,
       framebufferDimensions,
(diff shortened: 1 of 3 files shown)


src/core/renderers/webgl/WebGlCoreRenderer.ts
```diff
@@ -318,79 +319,112 @@ export class WebGlCoreRenderer extends CoreRenderer {
...
-
-      // Lower-Right
-      fQuadBuffer[bufferIdx++] = rightCornerX;
-      fQuadBuffer[bufferIdx++] = rightCornerY;
-      fQuadBuffer[bufferIdx++] = texCoordX2;
-      fQuadBuffer[bufferIdx++] = texCoordY2;
-      uiQuadBuffer[bufferIdx++] = colorBr;
-      fQuadBuffer[bufferIdx++] = textureIdx;
+      // render quad advanced
+      if (tb !== 0 || tc !== 0) {
```
frank-weindel, line 357: Could you move this conditional block up a level to be an `else if + else`? Just to reduce indentation where it isn't really needed

wouterlucas (approved):
nice!

src/core/renderers/webgl/WebGlCoreRenderer.ts
jfboeve: fixed

frank-weindel (approved):

erikhaandrikman (approved):

Issue #297, opened 2024-06-06
Wrong border radius on BorderEffect shader is cut


albdima83:
The view with border radius (50%) and border  width are cut.
See the picture attached:

<img width="376" alt="Screenshot 2024-06-06 alle 18 45 16" src="https://github.com/lightning-js/renderer/assets/62875440/1f9795b8-2263-4804-b27c-cb87b7b9545e">

Here a example code:

```javascript
const nodeCircleBorder = renderer.createNode({
    x: 100,
    y: 100,
    width: 152,
    height: 152,
    color: 0xff0000ff,
    shader: renderer.createShader('DynamicShader', {
      effects: [
        {
          type: 'radius',
          props: {
            radius: 152 / 2,
          },
        },
        {
          type: 'border',
          props: {
            width: 6,
            color: 0xfafafaff,
          },
        },
      ],
    }),
    parent: testRoot,
  });
```

pecoram:
@jfboeve

frank-weindel:
Fixed in #298

PR #298, opened 2024-06-10
Wrong border radius on BorderEffect shader is cut

albdima83:
Refer to issue: https://github.com/lightning-js/renderer/issues/297

Fix the cut border based on (https://github.com/rdkcentral/Lightning/blob/688237d7c705cedde0ad985c0a791047770c42fb/src/renderer/webgl/shaders/RoundedRectangleShader.mjs)

The problem is in BorderEffect.ts at this code that doesn't reach the last border pixel, see the picture below.

```javascript
float intR = shaderMask;
float mask = clamp(intR, 0.0, 1.0);
return mix(shaderColor, mix(shaderColor, maskColor, maskColor.a), mask);
```

Screenshot 2024-06-10 alle 11 31 45

After applying the fix, we reach the last border pixel adding one considering pixelRatio.
This solution is confirmed also in Lightning 2.0 (https://github.com/rdkcentral/Lightning/blob/688237d7c705cedde0ad985c0a791047770c42fb/src/renderer/webgl/shaders/RoundedRectangleShader.mjs)

```javascript
float intR = (shaderMask + 1.0 * u_pixelRatio);
float mask = clamp(intR, 0.0, 1.0);
return mix(shaderColor, mix(shaderColor, maskColor, maskColor.a), mask);
```

Screenshot 2024-06-10 alle 11 31 55


visual-regression/certified-snapshots/chromium-ci/dynamic-shader-circle-border-radius-1.png
frank-weindel, line 1: Great work! I'll let @jfboeve comment on the shader code specifically. But FYI there are other test snapshots that are failing that need to be updated in the PR for it to be accepted.

albdima83: @frank-weindel Thanks for your reply.
I added only the test *dynamic-shader-circle-border-radius-1*  but the github test failed, locally it works. 
Regarding the other tests, they were already failing before: you can test on main branch.

frank-weindel: @albdima83 I just ran the tests via `pnpm test:visual --ci` on main, and all tests are passing. That is a requirement for all PRs before they are merged to main. Are you making sure to run the tests in `--ci` mode when you run them locally?

frank-weindel: See `visual-regression/DOCKER.md` for more info

pecoram: dynamic-shader-1
this is what is generated on main. Maybe it is wrong

frank-weindel: @pecoram What computer model/processor/GPU was used to generate it? I have seen issues with the images generated with `--ci` when using a newer MacBook model with an M3 processor. Curious what you are using.

pecoram: MacBook with an M2 processor.. but the problem is that I replaced the dynamic-shader in the certified folder with new one because of our change. Shouldn't the workflow use the one I committed as the basis for comparison?

jfboeve (approved):

visual-regression/certified-snapshots/chromium-ci/dynamic-shader-circle-border-radius-1.png
frank-weindel: It was using yours and it was comparing it against one generated within GitHub Actions. There does seem to be an issue with the new Mac processors even when running the Chromium browser in Docker.

PR #299, opened 2024-06-12
refactor(CoreTextureManager): Eliminate ctxTextureCache

frank-weindel:
This refactor eliminates the ctxTextureCache WeakMap and replaces it with access to CoreContextTextures directly from Texture objects themselves.

Since access is more direct and does not require a map look up each time this change should be performance positive. It also simplifies the code a good bit.

This is being done as a foundational step for further Texture memory work.


wouterlucas (approved):
LGTM

PR #302, opened 2024-06-19
ShaderProps animation -> dev

jfboeve:
Added a new implementation for animating shaderProps.

Removed the shaderProps from the interface CoreNodeWritableProps because these are values that are always overwritten when setting a new Shader. 

I tried setting up a singular ShaderNode class that should have handled all the shader props, but that got too complicated. Currently its simplified to targeting the shaderProp property to animate the Shader.


erikhaandrikman:
Looks great! You only need to create a new snapshot for the test

erikhaandrikman (approved):

wouterlucas (approved):

jfboeve:
> Looks great! You only need to create a new snapshot for the test

Yeah I need to create a better snap shot anyway, will look into what we do at other animation tests

wouterlucas (approved):

frank-weindel:
Replaced by #314

PR #307, opened 2024-06-21
Add ability to query quad buffer usage info

marcel-danilewicz-consult-red:
Add quadBuffer usage log


elsassph:
That would be very noisy to console log on every frame. That doesn't seem very practical.

marcel-danilewicz-consult-red:
Good point. Every n frames? Let's say 60

elsassph:
I think it would be good to clarify what the objective of this PR is.

The code I see is not something you would put in a library - I mean, it's OK to put that in a branch of for your fork as a quick hack to get some stats for your own investigations, but that's not how it should be approached.

marcel-danilewicz-consult-red:
@elsassph Ok, could you clarify if either one is correct?
a) we need a permanent method to measure the quad buffer size in the future, but my solution is not good
b) only the result is important and the code should stay on a separate branch

elsassph:
Let's say that a & b are separate concerns. At the library level we could have a method to interrogate the renderer wrt the quad buffer usage, then at the application level you can read it and decide how to use it (console log, display like fps,...).

src/main-api/Renderer.ts
```diff
@@ -591,6 +591,10 @@ export class RendererMain extends EventEmitter {
     throw new Error('Not implemented');
   }
 
+  getQuadBufferUsage() {
```
elsassph, line 594: Probably not needed, as one can write the same code

src/core/renderers/webgl/WebGlCoreRenderer.ts
```diff
@@ -663,4 +666,7 @@ export class WebGlCoreRenderer extends CoreRenderer {
     }
     this.rttNodes.splice(index, 1);
   }
+  getQuadBufferUsage(): number {
```
elsassph, line 669: `quadBufferUsage` is public already

src/core/renderers/webgl/WebGlCoreRenderer.ts
```diff
@@ -87,6 +88,7 @@ export class WebGlCoreRenderer extends CoreRenderer {
    */
   defaultTexture: Texture;
 
+  quadBufferUsage = 0;
```
elsassph, line 91: Move up with the rest of `Persistent data`?

src/core/renderers/CoreRenderer.ts
```diff
@@ -102,4 +102,5 @@ export abstract class CoreRenderer {
   abstract renderRTTNodes(): void;
   abstract removeRTTNode(node: CoreNode): void;
   abstract renderToTexture(node: CoreNode): void;
+  abstract getQuadBufferUsage(): number | null;
```
elsassph, line 105: Maybe instead declare a public `quadBufferUsage?: number`

src/main-api/Renderer.ts
marcel-danilewicz-consult-red: But then usage would be 
`renderer.stage.renderer.quadBufferUsage`
instead of
`renderer.getQuadBufferUsage()`
I think, that the second one is a bit clearer, but the choice is yours

elsassph: Choice isn't mine :D but this is a fairly niche need so it's probably fine IMHO.

marcel-danilewicz-consult-red: Ok, let it stay  `renderer.stage.renderer.quadBufferUsage` then :)

frank-weindel:
Hey @marcel-danilewicz-consult-red @elsassph- Sorry I haven't been following this one too much. I think we should discuss this a bit if we're going to make this into a supported feature.

I recently implemented the ability to peak into texture memory with `renderer.stage.txMemManager.getMemoryInfo()` to support the MemMonitor that you can enable in the Test Examples. I thought of this mostly as an internal feature (or at least one that would never be used in production). Perhaps we could do something similar with `renderer.stage.renderer.getBufferInfo()`. And have it return an object which can be shaped different depending on the concrete CoreRenderer.

As a base type, `getBufferInfo()` could return:
```
{
  totalUsed: number;
  totalAvailable: number;
}
```
And have the numbers be reported in bytes, rather than index number.

Any thoughts?

marcel-danilewicz-consult-red:
Hi @frank-weindel :)
I've changed method name and return type. Using index number should be ok as the buffer type is ArrayBuffer which from what I understand is just an byte array. However I'm not sure about returning objects different than BufferInfo.

src/core/renderers/webgl/WebGlCoreRenderer.ts
```diff
@@ -694,6 +698,14 @@ export class WebGlCoreRenderer extends CoreRenderer {
     this.rttNodes.splice(index, 1);
   }
 
+  getBufferInfo(): BufferInfo | null {
+    const bufferInfo: BufferInfo = {
+      totalAvailable: QUAD_BUFFER_SIZE,
+      totalUsed: this.quadBufferUsage,
```
frank-weindel, line 704: This value is still just the last index. It should be the number of bytes used.

frank-weindel (approved):

PR #309, opened 2024-06-25
Fix: wrong load external resources when the local protocol is file://

albdima83:
When an external image resource is loaded from https, the function convertUrlToAbsolute inside the utils.ts doesn't check if it is a network or internal resource.

### DIFF
--- a/src/core/lib/utils.ts
+++ b/src/core/lib/utils.ts
@@ -252,8 +252,12 @@ export function isRectPositive(rect: Rect): boolean {
 }
 
 export function convertUrlToAbsolute(url: string): string {
-  // handle local file imports if the url is not the network resource
-  if (self.location.protocol === 'file:' && !EXTERNAL_REGEX_URL.test(url)) {
+  // handle local file imports if the url isn't remote resource or data blob
+  if (
+    self.location.protocol === 'file:' &&
+    !EXTERNAL_REGEX_URL.test(url) &&
+    !url.startsWith('data:')
+  ) {
     const path = self.location.pathname.split('/');
     path.pop();
     const basePath = path.join('/');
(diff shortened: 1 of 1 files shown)


src/core/lib/utils.ts
```diff
@@ -250,8 +252,12 @@ export function isRectPositive(rect: Rect): boolean {
 }
 
 export function convertUrlToAbsolute(url: string): string {
-  // handle local file imports
-  if (self.location.protocol === 'file:') {
+  // handle local file imports if the url isn't remote resource or data blob
+  if (
+    self.location.protocol === 'file:' &&
+    !EXTERNAL_REGEX_URL.test(url) &&
```
elsassph, line 258: Looks like all cases could be a single regex, if you just test `/^(data|https?|ftps?):/`

elsassph: And TBH, you could simplify further by just checking whether it looks like `{protocol}:`

albdima83: @elsassph Thanks, I fixed it! Simplify as `{protocol}:`

elsassph: Cool, I thought it might be simpler/safer to not test the protocol itself, e.g. any `^[a-z]+:` is a protocol.

albdima83: @elsassph I prefer to have a more specific check based on [protocol resources](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Identifying_resources_on_the_Web).
IMO your suggestion is too generic. 
What happen if you have an local image with filename like aaa:2222.png?

frank-weindel (approved):

wouterlucas (approved):
Good catch 😅 thanks! 😄

PR #313, opened 2024-06-27
Fix: Assign exact end values to Node properties when animations finish

### LINKED ISSUE #266: Issue with handling floating-point precision in node property animation
It seems to be an issue with handling floating-point values during node property animation. When animating a property in a node to a target float value, it's observed that instead of setting the exact target value upon animation completion, the node's property ends up with a nearby floating-point value that may have additional decimal places.

Observed this behaviour with most of the node proeprties. 

For example, if the initial value of an alpha property on a node is 1, and I animate it to 0.1, after the animation completes, the alpha value on the node might not be exactly 0.1, but rather something like 0.09999999991. 

This issue leads to some other issue in Blits transition

marcel-danilewicz-consult-red:
Add setting node properties to the end values after the animation is finished

### DIFF
--- a/src/core/animations/CoreAnimation.ts
+++ b/src/core/animations/CoreAnimation.ts
@@ -198,6 +198,11 @@ export class CoreAnimation extends EventEmitter {
         startValue + (endValue - startValue) * this.progress;
     }
     if (this.progress === 1) {
+      for (let i = 0; i < this.propsList.length; i++) {
+        const propName = this.propsList[i] as keyof CoreNodeAnimatableProps;
+        const propValue = this.props[propName] as number;
+        this.node[propName] = propValue;
+      }
       this.emit('finished', {});
     }
   }
--- a/src/core/animations/CoreAnimation.ts
+++ b/src/core/animations/CoreAnimation.ts
@@ -158,6 +158,11 @@ export class CoreAnimation extends EventEmitter {
       const startValue = this.propStartValues[propName] as number;
       const endValue = propValue;
 
+      if (this.progress === 1) {
+        this.node[propName] = endValue;
+        continue;
+      }
+
       if (propName.indexOf('color') !== -1) {
         // check if we have to change the color to begin with
         if (startValue === endValue) {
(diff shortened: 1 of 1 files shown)


src/core/animations/CoreAnimation.ts
```diff
@@ -198,6 +198,11 @@ export class CoreAnimation extends EventEmitter {
         startValue + (endValue - startValue) * this.progress;
     }
     if (this.progress === 1) {
+      for (let i = 0; i < this.propsList.length; i++) {
+        const propName = this.propsList[i] as keyof CoreNodeAnimatableProps;
+        const propValue = this.props[propName] as number;
+        this.node[propName] = propValue;
+      }
```
frank-weindel, line 205: Thanks! This should work fine but to squeeze maybe a tiny bit more performance out I think we could do a `progress === 1` check for line 197 and simply assign the `endValue` in that case.

marcel-danilewicz-consult-red: Do you mean something like this?
```
      if(this.progress===1){
        this.node[propName] = endValue;
      }
      else{
        this.node[propName] = startValue + (endValue - startValue) * this.progress;
      }
```
For some reason it doesn't work. Should I investigate further?

frank-weindel: Actually sorry I didn't even read the code close enough. You should be able to do it at the beginning of the for loop body:
```ts
if (this.progress === 1) {
  this.node[propName] = endValue;
  continue;
}
```

marcel-danilewicz-consult-red: Done

frank-weindel:
@marcel-danilewicz-consult-red This issue has been fixed. Rebase onto dev

marcel-danilewicz-consult-red:
It looks like this bug was also fixed in a commit, that is already present on dev, no need for rebasing and merging then.

frank-weindel:
@marcel-danilewicz-consult-red Ah yeah it looks like @jfboeve addressed this as part of his changes for Shader Animations #314

PR #314, opened 2024-06-27
Shader Animations w/ Dynamic Types for Shaders and Effects

jfboeve:
Added a new implementation for animating shaderProps.

Added ShaderController & DynamicShaderController functionalities. This allows us to change properties of shaders during runtime.

Optimised DynamicShader prop calculations, by having the DynamicShaderController do the calculations, which previously was done every time a node+dynamic shader where used in a renderOp.


examples/tests/shader-animation.ts
```diff
@@ -22,75 +22,125 @@ import type { ExampleSettings } from '../common/ExampleSettings.js';
...
-  //   if (t1.shader) {
-  //     t1.shader.radius = 10;
-  //   }
-  // });
+  shaderAnimation2.start();
+  await shaderAnimation2.waitUntilStopped();
+  t2Radius.text = 'radius: ' + t2.shader.props.r1.radius!.toString();
+  t2Border.text = 'border: ' + t2.shader.props.e1.width!.toString();
 
+  await snapshot({ name: 'animation2 finished' });
```
frank-weindel, line 144: Can you remove the space? This goes into the file name

src/core/CoreNode.ts
```diff
@@ -1079,7 +1062,7 @@ export class CoreNode extends EventEmitter {
       return false;
     }
 
-    if (this._shader) {
+    if (this.props.shader === this.stage.defShaderCtr) {
```
elsassph, line 1065: Should this be the opposite?

src/core/renderers/canvas/CanvasCoreRenderer.ts
```diff
@@ -54,6 +58,17 @@ export class CanvasCoreRenderer extends CoreRenderer {
...
     this.pixelRatio = pixelRatio;
     this.clearColor = clearColor ? getRgbaComponents(clearColor) : undefined;
+
+    // Stub for default shader controller since the canvas renderer does not
+    // (really) support the concept of a shader (yet)
+    this.defShaderCtr = {
+      type: 'DefaultShader',
+      props: {},
+      shader: new UnsupportedShader('DefaultShader'),
+      getResolvedProps: () => () => {
```
elsassph, line 68: That looks incorrect, and wasteful as it creates garbage on every invocation.

src/core/CoreNode.ts
```diff
@@ -676,18 +673,16 @@ export class CoreNode extends EventEmitter {
...
-  constructor(protected stage: Stage, props: CoreNodeWritableProps) {
+  constructor(readonly stage: Stage, props: CoreNodeProps) {
     super();
 
     this.props = {
       ...props,
       parent: null,
       texture: null,
-      shader: null,
+      shader: stage.defShaderCtr,
```
elsassph, line 685: Something inconsistent around shaders:
- custom shader is given through `props.shader`
- `this.props` has a spread of `props` but overrides `shader` with `stage.defShaderCtr`
- a few lines below `this.shader = props.shader`, and not `this.props.shader`
- rendering uses `this.shader`
- renderability doesn't test `this.shader`

src/core/CoreNode.ts
elsassph: Maybe 
```
if (this.shader !== this.stage.defShaderCtr)
```

src/core/CoreNode.ts
```diff
@@ -1833,33 +1812,20 @@ export class CoreNode extends EventEmitter {
...
-    this.loadShader(value.shType, value.props);
-  }
-
-  get shaderProps(): Record<string, unknown> | null {
-    return this.props.shaderProps;
-  }
 
-  set shaderProps(value: Record<string, unknown> | null) {
-    this.props.shaderProps = value;
+    if (value === this.stage.defShaderCtr) {
```
elsassph, line 1826: setting a custom shader doesn't change update type?

src/core/animations/CoreAnimation.ts
```diff
@@ -32,26 +32,69 @@ export interface AnimationSettings {
...
+          this.propValuesMap['shaderProps'][key] = {
+            start: node.shader.props[key] as number,
+            target: props.shaderProps[key] as number,
+          };
+        }
+      } else {
+        const shaderPropKeys = Object.keys(props.shaderProps!);
+        const spLength = shaderPropKeys.length;
+        let j = 0;
+        for (; j < spLength; j++) {
```
elsassph, line 80: why not `for (let j = 0;`?

src/core/animations/CoreAnimation.ts
```diff
@@ -152,51 +294,35 @@ export class CoreAnimation extends EventEmitter {
...
+        easing,
+      );
+    }
 
-      if (easing) {
-        this.node[propName] = this.applyEasing(
-          this.progress,
-          startValue,
-          endValue,
+    const dynEntries = Object.keys(this.dynPropValuesMap);
```
elsassph, line 312: Considering how rare the shader animations are going to be, it might have been more efficient to skip entirely this kind of code by having for instance `this.dynPropValuesMap` be `undefined` by default.

src/core/CoreNode.ts
frank-weindel: Not sure I understand

frank-weindel: Ah yes I see it now.

src/core/CoreNode.ts
frank-weindel: Yep not sure what I was thinking there.

src/core/CoreNode.ts
frank-weindel: `props` do already have default values assigned prior to CoreNode construction. The things happening in the constructor that you are calling out are to ensure that the setters are run for those particular properties. If the props weren't overridden in the spread then the setters would not actually run because of the "change check" at the top of each one.

But there is actually no reason at this point to do this for `shader` so I will remove it.

PR #315, opened 2024-06-27
fix: rtt in a parent with alpha 0

### LINKED ISSUE #312: RTT and parent alpha
When a child element with RTT (Render To Texture) is contained within a parent element that initially has an alpha value of 0, the child's texture does not update and remains invisible when the parent's alpha value is changed to 1.

Steps to Reproduce:

- Create a Parent element with an initial alpha value of 0.
- Add a child node to the Parent element. Ensure the child element has rtt set to true.
- Change the Parent element's alpha value from 0 to 1.

Expected Behavior:
The child element's texture should update and become visible when the Parent element's alpha value changes to 1.

Actual Behavior:
The child element's texture does not update and remains invisible, even though the Parent element's alpha value is set to 1.

pecoram:
fixes #312

- [x] visual tests

### DIFF
--- a/src/core/renderers/webgl/WebGlCoreRenderer.ts
+++ b/src/core/renderers/webgl/WebGlCoreRenderer.ts
@@ -605,8 +605,20 @@ export class WebGlCoreRenderer extends CoreRenderer {
       this.activeRttNode = node;
 
       assertTruthy(node.texture, 'RTT node missing texture');
+
+      if (node.texture.needsToBeRecreated === undefined && !node.isRenderable) {
+        node.texture.needsToBeRecreated = true;
+      } else if (
+        node.texture.needsToBeRecreated !== false &&
+        node.isRenderable
+      ) {
+        node.worldAlpha = 1;
+        node.texture.needsToBeRecreated = false;
+      }
+
       const ctxTexture = txManager.getCtxTexture(node.texture);
       assertTruthy(ctxTexture instanceof WebGlCoreCtxRenderTexture);
+
       this.renderToTextureActive = true;
 
       // Bind the the texture's framebuffer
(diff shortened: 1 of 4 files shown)


src/core/renderers/webgl/WebGlCoreRenderer.ts
```diff
@@ -605,8 +605,25 @@ export class WebGlCoreRenderer extends CoreRenderer {
...
+        !node.isRenderable
+      ) {
+        // The texture will have to be regenerated when the worldAlpha becomes 1
+        node.texture.needsToBeRegenerated = true;
+      } else if (
+        node.texture.needsToBeRegenerated !== false &&
+        node.isRenderable
+      ) {
+        // The texture must be regenerated with worldAlpha 1 because it was created with alpha 0
+        node.worldAlpha = 1;
```
frank-weindel, line 620: `worldAlpha` should only ever be updated by the `WorldAlpha` update type.

src/core/textures/Texture.ts
```diff
@@ -145,6 +145,8 @@ export abstract class Texture extends EventEmitter {
 
   readonly renderableOwners = new Set<unknown>();
 
+  needsToBeRegenerated: boolean | undefined = undefined;
```
frank-weindel, line 148: Not a fan of adding this to the general `Texture` since it is only used for RTT. There is RenderTexture however I'm curious what @erikhaandrikman's thoughts are.

frank-weindel:
I'm wondering if there's a way that maybe the code that renders the texture could be somehow implemented in the `getTextureData()` function of RenderTexture. Then it could automatically benefit from renderability changes in terms being rendered initially when becoming renderable and also potentially being freed during texture cleanup when unrenderable.

wouterlucas:
Replaced with https://github.com/lightning-js/renderer/pull/394

Issue #283, opened 2024-05-29
Clamp the radius on the RadiusEffect shader


WillsonHaw:
The radius value for the RadiusEffect shader is currently unbounded and can be set to a really high value. 

Here's how it looks in Lightning v3 with a radius of 55:

![image](https://github.com/lightning-js/renderer/assets/600962/1e4e2753-24b6-4ea9-933f-177564f9ee9d)

Here's how it looks in Web with the same values:
![image](https://github.com/lightning-js/renderer/assets/600962/ade69c3c-69d1-41e2-8301-76e7bcd381ce)

M4tiz:
Looks like normalization of overlapping curves should be applied here. (specified by [CSS 3 spec](https://drafts.csswg.org/css-backgrounds/#corner-overlap) ). I can handle that :)

wouterlucas:
Solved in #317

PR #317, opened 2024-07-01
Clamp the radius on the RoundedRectangle / RadiusEffect shader

M4tiz:
Following the CSS3 spec
https://drafts.csswg.org/css-backgrounds-3/#corner-overlap

### DIFF
--- a/src/core/renderers/webgl/shaders/RoundedRectangle.ts
+++ b/src/core/renderers/webgl/shaders/RoundedRectangle.ts
@@ -77,7 +77,10 @@ export class RoundedRectangle extends WebGlCoreShader {
   }
 
   protected override bindProps(props: Required<RoundedRectangleProps>): void {
-    this.setUniform('u_radius', props.radius);
+    const radiusFactor =
+      Math.min(props.$dimensions.width, props.$dimensions.height) /
+      (2.0 * props.radius);
+    this.setUniform('u_radius', props.radius * Math.min(radiusFactor, 1));
   }
 
   override canBatchShaderProps(


M4tiz:
Fix for #283

M4tiz:
Updated `destroy-1` snapshot since the change affects `destroy.ts` nodes - they have border radius greater than allowed which gets clamped.

src/core/renderers/webgl/shaders/RoundedRectangle.ts
```diff
@@ -119,6 +122,8 @@ export class RoundedRectangle extends WebGlCoreShader {
         // pass to fragment
         v_color = a_color;
         v_textureCoordinate = a_textureCoordinate;
+        float radiusFactor = min(u_dimensions.x, u_dimensions.y) / (2.0 * u_radius);
+        v_radius = u_radius * min(radiusFactor, 1.0);
```
frank-weindel, line 126: I think this calculation can be done in the CPU bound code so that it only has to run once per shader invocation rather than for each pixel. Same for Effect I imagine.

M4tiz: I managed to do this for `RoundedRectangle` however it's more complicated in `RadiusEffect` since in the effect instance object we don't have access to parent Shader's dimensions. Any hints on how to address that?

frank-weindel: @jfboeve Any ideas?

M4tiz: @frank-weindel I made a possible solution. It requires passing `DynamicShaderProps` down to `RadiusEffect` uniform radius' validator function. It's quite a simple change, however it adds contract dependency to `DynamicShader` in `ShaderEffect`. Let me know what do you think :)

frank-weindel: Looks simple enough to me! @jfboeve?

jfboeve: For the normal RoundedRectangle it looks okay. for the RadiusEffect also, however that only works for now with the current dev branch without the ShaderController/DynamicShaderController that will be implemented soon. With that release the validation part is moved to cut back on unnecessary calculations. This means that the solution would work if the nodes width/height would not be changed during runtime, however it would not properly clamp when the width/height are animated while the radius is not being changed... 

I've a few ideas to resolve this problem where the effectprops need to be recalculated based on the automatic values width/height/alpha.

frank-weindel:
Looks like @jfboeve addressed this in #314 with some other refactorings.

jfboeve:
> Looks like @jfboeve addressed this in #314 with some other refactorings.

Corrrect! Since I needed an effect to test if fixed the Radius clamping using the solution used here, so only the RoundedRectangle fix should be pushed.

src/core/renderers/webgl/shaders/RoundedRectangle.ts
M4tiz: Reverted the `RadiusEffect` changes :)

frank-weindel (approved):

PR #319, opened 2024-07-02
Improve Canvas Text Renderer Performance

frank-weindel:
- Simplify/improve canvas renderer performance by utilizing child nodes and ImageTexture factory.
- All canvas text is now laid out when created, but not actually rasterized until the text becomes renderable (inbounds, positive alpha, etc)
- Canvas textures are now properly able to be cleaned up when they become unrenderable.

# TODO
- [x] Run performance tests
- [x] ~Get multi-page scrolling working again (or this can be deferred)~ #321


frank-weindel:
@wouterlucas Could you help me performance test this if you have a chance?

src/core/Stage.ts
```diff
@@ -448,4 +453,108 @@ export class Stage {
     // the covariant state argument in the setter method map
     return resolvedTextRenderer as unknown as TextRenderer;
   }
+
+  createNode(props: Partial<CoreNodeWritableProps>) {
+    const resolvedProps = this.resolveNodeDefaults(props);
+    const node = new CoreNode(this, {
```
elsassph, line 459: Seems a waste to clone the `resolvedProps` just to add `shaderProps: null`

src/core/text-rendering/renderers/CanvasTextRenderer.ts
```diff
@@ -369,398 +319,145 @@ export class CanvasTextRenderer extends TextRenderer<CanvasTextRendererState> {
...
+          this.canvas.height,
+        );
+      }.bind(this, state.lightning2TextRenderer, state.renderInfo),
+    });
+    if (state.textureNode) {
+      // Use the existing texture node
+      state.textureNode.texture = texture;
+    } else {
+      // Create a new texture node
+      const textureNode = this.stage.createNode({
```
elsassph, line 368: Why creating a sub node? Is that to decouple the text node size and the texture?

src/core/Stage.ts
frank-weindel: Agreed. Will fix.

wouterlucas:
> @wouterlucas Could you help me performance test this if you have a chance?



> @wouterlucas Could you help me performance test this if you have a chance?

See here:
image

All 3 renderer using Canvas fonts. If you want I can add the same codebase that the PR is from and enable MSDF fonts for comparison. Though it will be fastest.

But still a bit away from L2's performance.

Issue #316, opened 2024-07-01
CoreNode: Only create scaleRotateTransform matrix when needed


frank-weindel:
On a CoreNode, if it's `rotation` prop is 0 and both it's `scaleX` and `scaleY` props are 1, there is no need to create a `scaleRotateTransform`. In `updateLocalTransform` we can check if the `scaleRotateTransform` exists, and if not, simply create the Local Transform without using it. The `pivotX` and `pivotY` properties can also be completely ignored in this case since they only apply during scaling/rotation.

Since likely most Nodes in applications will not be need scaling or rotation this should save some CPU bound memory per node while also cutting down on calculations done during the first update for Nodes.

PR #320, opened 2024-07-02
Only create scaleRotateTransform matrix when needed

elsassph:
Implements #316

`scaleRotateTransform` matrix is a waste of memory and computations when `rotation=0` and `scale=1`, which will be the case 99.9% of the time in an average app.


frank-weindel (approved):
LGTM. Thanks!

Issue #318, opened 2024-07-02
Animations with stopMethod 'reverse' never get cleaned up from AnimationManager


frank-weindel:
As the title says. The animations stay in the active animations Set even after the reverse animation completes.

wouterlucas:
Solved in #339

PR #339, opened 2024-07-15
Fix: Remove from active animations when reverse stopMethod is complete

Drulokia:
#318 - After stopMethod 'reverse' is done it is removed from animations. Also the animation loops in its entirety not only the reverse part when loop is true.


wouterlucas (approved):
LGTM

PR #344, opened 2024-07-17
Fix for error caused by default font

Drulokia:


src/core/text-rendering/font-face-types/WebTrFontFace.ts
```diff
@@ -67,15 +67,17 @@ export class WebTrFontFace extends TrFontFace {
...
-      })
-      .catch(console.error);
-
+    // Temperary Fix for error in FontFace.load() when fontUrl is empty caused by default renderer font
+    if (fontUrlWithoutParentheses.length > 0) {
+      fontFace
+        .load()
+        .then(() => {
+          // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+          (this.loaded as boolean) = true;
```
elsassph, line 76: For consistency we should set the `loaded` flag in the `else` (and probably `catch`) case.

pecoram: I don't know if it is useful to know when the font has gone wrong. Maybe in that case emit("errored")

elsassph: Yes, I guess if you touch error handling it could emit an `errored` message, but maybe that's scope creep.

Drulokia: Cannot set the loaded flag outside the promise of a succesfull load.

elsassph: Of course you can

src/core/text-rendering/font-face-types/WebTrFontFace.ts
```diff
@@ -74,7 +74,13 @@ export class WebTrFontFace extends TrFontFace {
         (this.loaded as boolean) = true;
         this.emit('loaded');
       })
-      .catch(console.error);
+      .catch((error) => {
```
elsassph, line 77: We still want to avoid loading the font at all when it's an empty string.
e.g. `if (...) {} else { (this.loaded as boolean) = true; }`

wouterlucas (approved):
LGTM

elsassph:
@wouterlucas @Drulokia little mistake: in the `else` case, you shouldn't invoke `this.emit('loaded');` because as it runs synchronously in the constructor it's not possible to have added a listener.

Issue #268, opened 2024-05-03
Wrong wrapping with contain width and text too long


pecoram:
With `contain: "width"` a long text is not wrapped correctly:

![image](https://github.com/lightning-js/renderer/assets/33352553/f703db99-6c19-4ca4-a9d2-d8b973a2f9a2)

this issue is for both canvas and sdf

marcel-danilewicz-consult-red:
Just to clarify - is `contain: "width"` equivalent of css `text-wrap: wrap`? If yes, then I think, that this behavior may be correct. I checked https://developer.mozilla.org/en-US/docs/Web/CSS/text-wrap and it behaves similarly. The text overflows only if it is continuous, without whitespaces.

pecoram:
> Just to clarify - is `contain: "width"` equivalent of css `text-wrap: wrap`? If yes, then I think, that this behavior may be correct. I checked https://developer.mozilla.org/en-US/docs/Web/CSS/text-wrap and it behaves similarly. The text overflows only if it is continuous, without whitespaces.

`contain: "width"` must behave as `text-wrap: wrap` so if a word is too long the excess part must wrap. Just like the example on the page you linked

marcel-danilewicz-consult-red:
I was able to achieve the same result as in lightning3:
![image](https://github.com/lightning-js/renderer/assets/172387148/84c98ad9-e163-423d-94a3-4fd9d7ef1201)

marcel-danilewicz-consult-red:
In this case should this ticket be treated as invalid or maybe we want to implement the wrap-word functionality in it's scope?

marcel-danilewicz-consult-red:
I tried tinkering with the code in case we wanted to implement this. Is this the expected behavior for "wrap-word" or should the overflow at the bottom not happen?
![image](https://github.com/lightning-js/renderer/assets/172387148/2a0006f0-e1e2-4237-bd76-67c22e397096)

pecoram:
In my opinion it is wrong to wrap with spaces but if we want we can add a separate parameter that we call wordWrap and if it is true it truncates the word and not the last space

marcel-danilewicz-consult-red:
Hi @wouterlucas 
I'd like to clarify the requirements for this functionality. I see two solutions:
a) any word, that exceeds the width limit should wrap
b) the word, that exceeds the width limit should be moved to the next line. If it stil exceeds the limit it should wrap until it fits
Is either one correct? Are there any other requirements?

[...]

PR #345, opened 2024-07-17
Implement wrapWord parameter (#268)

marcel-danilewicz-consult-red:
WiP, todo: canvas renderer

### DIFF
--- a/src/core/text-rendering/renderers/LightningTextTextureRenderer.ts
+++ b/src/core/text-rendering/renderers/LightningTextTextureRenderer.ts
@@ -699,7 +699,7 @@ export class LightningTextTextureRenderer {
         const wordWidth = this.measureText(words[j]!, letterSpacing);
         const wordWidthWithSpace =
           wordWidth + this.measureText(' ', letterSpacing);
-        if (wordWidthWithSpace > wordWrapWidth && wordBreak) {
+        if (wordBreak && wordWidthWithSpace > wordWrapWidth) {
           let remainder = '';
           while (this.measureText(words[j]!) > wordWrapWidth) {
             remainder = words[j]!.slice(-1) + remainder;
--- a/src/core/text-rendering/renderers/LightningTextTextureRenderer.ts
+++ b/src/core/text-rendering/renderers/LightningTextTextureRenderer.ts
@@ -323,6 +323,7 @@ export class LightningTextTextureRenderer {
         wordWrapWidth,
         letterSpacing,
         textIndent,
+        this._settings.wordBreak,
       );
     } else {
       linesInfo = { l: this._settings.text.split(/(?:\r\n|\r|\n)/), n: [] };
@@ -347,6 +348,7 @@ export class LightningTextTextureRenderer {
           wordWrapWidth - w,
           letterSpacing,
           textIndent,
+          this._settings.wordBreak,
         );
         usedLines[usedLines.length - 1] = `${al.l[0]!}${
           this._settings.overflowSuffix
@@ -681,6 +683,7 @@ export class LightningTextTextureRenderer {
     wordWrapWidth: number,
     letterSpacing: number,
     indent = 0,
+    wordBreak: boolean,
   ) {
     // Greedy wrapping algorithm that will wrap words as the line grows longer.
     // than its horizontal bounds.
@@ -718,6 +721,18 @@ export class LightningTextTextureRenderer {
 
       if (i < lines.length - 1) {
         realNewlines.push(allLines.length);
+      }
+    }
+    if (wordBreak) {
+      for (let i = 0; i < allLines.length; i++) {
+        let newline = '';
+        while (this.measureText(allLines[i]!, letterSpacing) > wordWrapWidth) {
+          newline = allLines[i]!.slice(-1) + newline;
+          allLines[i] = allLines[i]!.slice(0, -1);
+        }
+        if (newline.length > 0) {
+          allLines.splice(i + 1, 0, newline);
+        }
       }
     }
 
--- a/src/core/text-rendering/renderers/LightningTextTextureRenderer.ts
+++ b/src/core/text-rendering/renderers/LightningTextTextureRenderer.ts
@@ -699,6 +699,16 @@ export class LightningTextTextureRenderer {
         const wordWidth = this.measureText(words[j]!, letterSpacing);
         const wordWidthWithSpace =
           wordWidth + this.measureText(' ', letterSpacing);
+        if (wordWidthWithSpace > wordWrapWidth && wordBreak) {
+          let remainder = '';
+          while (this.measureText(words[j]!) > wordWrapWidth) {
+            remainder = words[j]!.slice(-1) + remainder;
+            words[j] = words[j]!.slice(0, -1);
+          }
+          if (remainder.length > 0) {
+            words.splice(j + 1, 0, remainder);
+          }
+        }
         if (j === 0 || wordWidthWithSpace > spaceLeft) {
           // Skip printing the newline if it's the first word of the line that is.
           // greater than the word wrap width.
@@ -723,19 +733,6 @@ export class LightningTextTextureRenderer {
         realNewlines.push(allLines.length);
       }
     }
-    if (wordBreak) {
-      for (let i = 0; i < allLines.length; i++) {
-        let newline = '';
-        while (this.measureText(allLines[i]!, letterSpacing) > wordWrapWidth) {
-          newline = allLines[i]!.slice(-1) + newline;
-          allLines[i] = allLines[i]!.slice(0, -1);
-        }
-        if (newline.length > 0) {
-          allLines.splice(i + 1, 0, newline);
-        }
-      }
-    }
-
     return { l: allLines, n: realNewlines };
   }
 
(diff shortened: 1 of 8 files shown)


src/core/text-rendering/renderers/SdfTextRenderer/internal/layoutText.ts
```diff
@@ -292,6 +293,24 @@ export function layoutText(
           maxX = Math.max(maxX, quadX + glyph.width);
           curX += glyph.xAdvance;
         }
+        if (
+          wrapWord &&
+          charEndX + glyph.width >= lineVertexW &&
```
elsassph, line 298: Is that correct? Wrap-word I think is meant to only wrap a word when this word alone is larger than the wrapping width.

marcel-danilewicz-consult-red: I thought it should be wrapped if the width is exceeded. So if a word doesn't fit it should be moved to the next line and then if still doesn't fit it should be wrapped?

elsassph: I think it would be silly to break every word. The point is that sometimes (as reported in #268) you may have a single word larger than the target width, in which case you may prefer to break it. 

In L2, that's what `wordBreak` option does. I recommend checking the implementation (which is TBH rather inefficient). The algorithm actually breaks words "before" doing the layout, so for instance a very long word would be always cut to not exceed the full width, but it won't try to fit a few characters on the previous line. It also supports words being multiple times longer than the available width - imagine for instance using this feature to render text vertically, one letter per line.

I suggest again to make sure that the requirements are non-ambiguous - ask @wouterlucas for confirmation. Make a diagram/drawing to explain what you are going to do.

In any case it is very wise to have proposed an early draft.

src/core/text-rendering/renderers/LightningTextTextureRenderer.ts
```diff
@@ -720,6 +723,18 @@ export class LightningTextTextureRenderer {
         realNewlines.push(allLines.length);
       }
     }
+    if (wordBreak) {
```
elsassph, line 726: I think this is incorrect in some cases, and inefficient in general. You should instead break words as part of the `wrapText` function.

marcel-danilewicz-consult-red: But this is inside of `wrapText` function :) 
Could you give me an example of NG case for this code?

elsassph: Right, but it is misplaced I think: it re-measures every word which we already measure in the above loop. Also isn't that a different breaking logic as SDF? It seems SDF tries to cut the word to fit the line, isn't it?

marcel-danilewicz-consult-red: It rather remeasures whole lines. I'd rather separate normal and break cases, but you are right that the code can be optimized. Algorithms are not the same due to different surrounding code.

src/core/text-rendering/renderers/LightningTextTextureRenderer.ts
```diff
@@ -696,6 +699,16 @@ export class LightningTextTextureRenderer {
...
+          while (this.measureText(words[j]!) > wordWrapWidth) {
+            remainder = words[j]!.slice(-1) + remainder;
+            words[j] = words[j]!.slice(0, -1);
+          }
+          if (remainder.length > 0) {
+            words.splice(j + 1, 0, remainder);
+          }
+        }
         if (j === 0 || wordWidthWithSpace > spaceLeft) {
           // Skip printing the newline if it's the first word of the line that is.
```
elsassph, line 713: I think this is where you cut the word to fix within the `spaceLeft`, and then what remains is cut if remainder still goes over `wordWrapWidth`.

marcel-danilewicz-consult-red: Yes, but what remains is cut in next iteration of `for (let j = 0; j < words.length; j++)` loop as it is inserted to `words` at `j+1` index and thus treated as another word.

elsassph: Ok it can work here

src/core/text-rendering/renderers/SdfTextRenderer/internal/layoutText.ts
```diff
@@ -292,6 +293,24 @@ export function layoutText(
           maxX = Math.max(maxX, quadX + glyph.width);
           curX += glyph.xAdvance;
         }
+        if (
```
elsassph, line 296: Same comment about order as above: 
```
if (wrapWord == 'break' && contain != 'none' && the rest)
```

src/core/text-rendering/renderers/LightningTextTextureRenderer.ts
```diff
@@ -696,6 +699,16 @@ export class LightningTextTextureRenderer {
         const wordWidth = this.measureText(words[j]!, letterSpacing);
         const wordWidthWithSpace =
           wordWidth + this.measureText(' ', letterSpacing);
+        if (wordWidthWithSpace > wordWrapWidth && wordBreak) {
```
elsassph, line 702: Put conditions in narrowing order:
```
if (wordBreak && the rest) {
```

elsassph, line 702: I think the algorithm here is different from SDF.

Isn't the goal agreed to cut the word so the beginning of the word fits at the end of the line? e.g. keep what fits within `spaceLeft` and move the remainder to the next line (where indeed it could still be `> wordWrapWidth`)

src/core/text-rendering/renderers/LightningTextTextureRenderer.ts
elsassph: BTW I know that this project tends to use mostly visual regression tests, but this is a case where some unit tests would be greatly beneficial @wouterlucas

src/core/text-rendering/renderers/LightningTextTextureRenderer.ts
```diff
@@ -691,11 +694,21 @@ export class LightningTextTextureRenderer {
       const resultLines = [];
       let result = '';
       let spaceLeft = wordWrapWidth - indent;
-      const words = lines[i]!.split(' ');
+      const words = wordBreak ? [lines[i]!] : lines[i]!.split(' ');
```
elsassph, line 697: Ok that's kind of right and wrong - very interesting.

The good: 
- If we're breaking text we don't really care about spaces anymore; we can "just" find how many characters fit, and fill the lines,

The bad:
- This is very inefficient,
- It awkwardly integrates with the existing loop.

Suggestion:
```typescript
if (wordBreak) {
  // approach breaking the lines
} else {
  // original code splitting and wrapping the words
}
```

Performant approach:
- This class has a `wrapWord(word, wordWrapWidth, suffix)` method which has an optimized algorithm to cut a string at a max width. It will return the part that fits, from which you can figure the remainder and repeat until everything has fit.

marcel-danilewicz-consult-red: Good point. I tried the wrapWord, but the line break points are a bit different between sdf and canvas.
image
image
I'll push the commit if this is ok.

elsassph: Hmm ok curious. Why is `sit` on the next line for SDF? And also why did it keep the space character when wrapping `ex magna`?
Aside from the the canvas wrapping looks alright, if the definition is to fit as many characters per line as possible.

marcel-danilewicz-consult-red: I got stuck a little bit on this issue. To prevent `sit` from being carried to the next line additional `wrapWord=='normal'` condition is needed in line 213:
```
 // Word wrap check
        if (
          // We are containing the text
          contain !== 'none' &&
          wrapWord == 'normal' &&
```

Then to break at the correct point
`charEndX + glyph.width >= lineVertexW `(line 299)
should be changed to
`charEndX + glyph.width + glyph.xOffset + 4 >= lineVertexW`

The value of 4 was concluded experimentally and I'm not even sure why it is needed there. Here you can see the how breaking looks without it:
image
and with:
image

Finally - I have no idea about additional space before "ex".

elsassph: About `sit`, I guess we should either wrap or break works, no?

Adding +4 isn't worth it - it's just the subtle differences in how text layout happens.

For the space before `ex` it's probably because when you cut the line, it cuts just at the space characters, so line starts with a space. We shouldn't carry over a space character to next line when we break, so probably we can remove a starting space character at the beginning of a wrapped line.

src/core/text-rendering/renderers/LightningTextTextureRenderer.ts
wouterlucas: > BTW I know that this project tends to use mostly visual regression tests, but this is a case where some unit tests would be greatly beneficial @wouterlucas

theres a vitest infrastructure as well, just add `*.test.ts` and it should be picked up

pecoram:
What is missing to complete this PR?

wouterlucas:
PR is stale and fixed partially with #522 without introducing new parameters.

PR #348, opened 2024-07-24
Support for Tree Shakable Canvas/WebGL renderer and Font Engines

wouterlucas:
* Expose Canvas and WebGL Core Renderers under `@lightningjs/renderer/webgl` and `@lightningjs/renderer/canvas` respectively
* Allow RendererMain config to accept a `renderEngine` parameter where the core renderers can be provided
* Expose SDF Text Renderer and CanvasTextRenderer under canvas/webgl exports.
* Allow RendererMain config to accept an array of `fontEngines` where you can provide 1 or 2 Font Engines.
* Fallback with Canvas still works IF Canvas Text Renderer is provided
* Drop memize dependency to reduce overall bundle size, apply simple caching instead

Todo:
- [x] Check tests, see if text renderer override still works
- [x] Run through some of the resolve code again to make sure we're not introducing new bottlenecks

In my tests the simple benchmark app went down by `20 kb` by just using 1 font engine to `40 kb` when only using canvas and canvas rendering.

Note: This is a breaking init change, handle with care.

### DIFF
--- a/src/core/CoreTextNode.ts
+++ b/src/core/CoreTextNode.ts
@@ -35,7 +35,6 @@ import type {
 import type { RectWithValid } from './lib/utils.js';
 import { assertTruthy } from '../utils.js';
 import { Matrix3d } from './lib/Matrix3d.js';
-import { EventEmitter } from '../common/EventEmitter.js';
 
 export interface CoreTextNodeProps extends CoreNodeProps, TrProps {
   /**
@@ -71,43 +70,46 @@ export interface CoreTextNodeProps extends CoreNodeProps, TrProps {
  * For non-text rendering, see {@link CoreNode}.
  */
 export class CoreTextNode extends CoreNode implements CoreTextNodeProps {
-  textRenderer: TextRenderer | null = null;
+  textRenderer: TextRenderer;
   trState: TextRendererState;
   private _textRendererOverride: CoreTextNodeProps['textRendererOverride'] =
     null;
 
-  constructor(stage: Stage, props: CoreTextNodeProps) {
+  constructor(
+    stage: Stage,
+    props: CoreTextNodeProps,
+    textRenderer: TextRenderer,
+  ) {
     super(stage, props);
     this._textRendererOverride = props.textRendererOverride;
-    const { resolvedTextRenderer, textRendererState } =
-      this.resolveTextRendererAndState({
-        x: this.absX,
-        y: this.absY,
-        width: props.width,
-        height: props.height,
-        textAlign: props.textAlign,
-        color: props.color,
-        zIndex: props.zIndex,
-        contain: props.contain,
-        scrollable: props.scrollable,
-        scrollY: props.scrollY,
-        offsetY: props.offsetY,
-        letterSpacing: props.letterSpacing,
-        debug: props.debug,
-        fontFamily: props.fontFamily,
-        fontSize: props.fontSize,
-        fontStretch: props.fontStretch,
-        fontStyle: props.fontStyle,
-        fontWeight: props.fontWeight,
-        text: props.text,
-        lineHeight: props.lineHeight,
-        maxLines: props.maxLines,
-        textBaseline: props.textBaseline,
-        verticalAlign: props.verticalAlign,
-        overflowSuffix: props.overflowSuffix,
-      });
+    this.textRenderer = textRenderer;
+    const textRendererState = this.createState({
+      x: this.absX,
+      y: this.absY,
+      width: props.width,
+      height: props.height,
+      textAlign: props.textAlign,
+      color: props.color,
+      zIndex: props.zIndex,
+      contain: props.contain,
+      scrollable: props.scrollable,
+      scrollY: props.scrollY,
+      offsetY: props.offsetY,
+      letterSpacing: props.letterSpacing,
+      debug: props.debug,
+      fontFamily: props.fontFamily,
+      fontSize: props.fontSize,
+      fontStretch: props.fontStretch,
+      fontStyle: props.fontStyle,
+      fontWeight: props.fontWeight,
+      text: props.text,
+      lineHeight: props.lineHeight,
+      maxLines: props.maxLines,
+      textBaseline: props.textBaseline,
+      verticalAlign: props.verticalAlign,
+      overflowSuffix: props.overflowSuffix,
+    });
 
-    this.textRenderer = resolvedTextRenderer;
     this.trState = textRendererState;
   }
 
@@ -155,7 +157,7 @@ export class CoreTextNode extends CoreNode implements CoreTextNodeProps {
 
   override set width(value: number) {
     this.props.width = value;
-    this.textRenderer?.set.width(this.trState, value);
+    this.textRenderer.set.width(this.trState, value);
 
     // If not containing, we must update the local transform to account for the
     // new width
@@ -170,7 +172,7 @@ export class CoreTextNode extends CoreNode implements CoreTextNodeProps {
 
   override set height(value: number) {
     this.props.height = value;
-    this.textRenderer?.set.height(this.trState, value);
+    this.textRenderer.set.height(this.trState, value);
 
     // If not containing in the horizontal direction, we must update the local
     // transform to account for the new height
@@ -184,7 +186,7 @@ export class CoreTextNode extends CoreNode implements CoreTextNodeProps {
   }
 
   override set color(value: number) {
-    this.textRenderer?.set.color(this.trState, value);
+    this.textRenderer.set.color(this.trState, value);
   }
 
   get text(): string {
@@ -192,7 +194,7 @@ export class CoreTextNode extends CoreNode implements CoreTextNodeProps {
   }
 
   set text(value: string) {
-    this.textRenderer?.set.text(this.trState, value);
+    this.textRenderer.set.text(this.trState, value);
   }
 
   get textRendererOverride(): CoreTextNodeProps['textRendererOverride'] {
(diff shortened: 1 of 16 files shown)


src/core/CoreTextNode.ts
```diff
@@ -153,7 +155,7 @@ export class CoreTextNode extends CoreNode implements CoreTextNodeProps {
 
   override set width(value: number) {
     this.props.width = value;
-    this.textRenderer.set.width(this.trState, value);
+    this.textRenderer?.set.width(this.trState, value);
```
chiefcll, line 158: Since the textRenderer is set in the constructor you shouldn't need the ? for any of these.

src/core/CoreTextNode.ts
```diff
@@ -70,7 +71,7 @@ export interface CoreTextNodeProps extends CoreNodeProps, TrProps {
  * For non-text rendering, see {@link CoreNode}.
  */
 export class CoreTextNode extends CoreNode implements CoreTextNodeProps {
-  textRenderer: TextRenderer;
+  textRenderer: TextRenderer | null = null;
```
chiefcll, line 74: This is set in the constructor - I'd assume you'd want a textRenderer? Making this allow nulls causes all those ?

src/core/Stage.ts
```diff
@@ -62,9 +61,10 @@ export interface StageOptions {
   fpsUpdateInterval: number;
   enableContextSpy: boolean;
   numImageWorkers: number;
-  renderMode: 'webgl' | 'canvas';
+  renderEngine: typeof WebGlCoreRenderer | typeof CanvasCoreRenderer;
```
chiefcll, line 64: do you need the typeof? Those are types - `WebGlCoreRenderer | CanvasCoreRenderer`

src/core/Stage.ts
```diff
@@ -62,9 +61,10 @@ export interface StageOptions {
...
   eventBus: EventEmitter;
   quadBufferSize: number;
+  fontEngines: (typeof CanvasTextRenderer | typeof SdfTextRenderer)[];
```
chiefcll, line 67: same

src/main-api/Renderer.ts
```diff
@@ -193,16 +197,53 @@ export interface RendererMainSettings {
...
+   * WebGL is more performant and supports more features. Canvas is
+   * supported on most platforms.
+   *
+   * Note: When using CanvasCoreRenderer you can only use
+   * CanvasTextRenderer. The WebGLCoreRenderer supports
+   * both CanvasTextRenderer and SdfTextRenderer for Text Rendering.
+   *
    */
-  renderMode?: 'webgl' | 'canvas';
+  renderEngine: typeof CanvasCoreRenderer | typeof WebGlCoreRenderer;
```
chiefcll, line 212: CanvasCoreRenderer | WebGlCoreRenderer

src/core/CoreTextNode.ts
wouterlucas: with the Text Renderers being provided as a setting, there is a possibility there is no text renderer provided at all. Hence it can be `null` and needs to be checked before calling.

Granted this whole getters/setters calling getters/setters is a little redundant and does not help performance, could be optimized but that's for another PR.

src/core/CoreTextNode.ts
wouterlucas: Yes - since you can provide no fontEngine at all on Renderer init, this can be `null`.

src/core/Stage.ts
wouterlucas: Yes TypeScript _requires_ the `typeof` here.

The difference is an instance of the class `WebGlCoreRenderer` / `CanvasCoreRenderer` versus the constructable class `WebGlCoreRenderer` / `CanvasCoreRenderer`. Since we're doing the `new` instance creation of those renderers in the Stage constructor, we have to pass the Class itself and not an instance of the class. That requires `typeof` in TypeScript.

src/core/Stage.ts
wouterlucas: See above.

src/main-api/Renderer.ts
wouterlucas: See above, needs to be `typeof`.

src/core/CoreTextNode.ts
chiefcll: I get that it can be Null in the case the user doesn't specify a Text Renderer and Creates a text node. To avoid all the null checks you could have single check when creating a CoreTextNode. All the ? would be ! to tell TS you know better and no performance overhead

src/core/CoreTextNode.ts
elsassph: Not sure it's worth adding nullability in the mix: if you try to use a text node but don't have any text renderer, throw.

src/core/CoreTextNode.ts
```diff
@@ -168,7 +170,7 @@ export class CoreTextNode extends CoreNode implements CoreTextNodeProps {
 
   override set height(value: number) {
     this.props.height = value;
-    this.textRenderer.set.height(this.trState, value);
+    this.textRenderer?.set.height(this.trState, value);
```
elsassph, line 173: @wouterlucas, as @chiefcll said, this is useless adding runtime overhead - this node should fail (early) if you don't have a text renderer.

src/core/Stage.ts
elsassph, line 64: Why not type as `CoreRenderer`?

src/core/CoreTextNode.ts
wouterlucas: Yeah I like your idea of throwing if that happens, that avoids dealing with anything else on the CoreTexTNode.

src/core/Stage.ts
wouterlucas: Sure that works too, does it really matter? It's just typescript telling typescript what to adhere too (i.e. doesn't change the behavior nor affect type safety).

src/core/CoreTextNode.ts
wouterlucas: Okay so quick observations with this, it's not that trivial. The getter/setter checks are relatively inexpensive as they do not get called that frequently, moving checks to another place means we have to deal with the renderer finding the `CoreTextNode` as its walking the tree and ask for its renderability.

Prior to this change the Renderer would always be able to fall back to CanvasTextRenderer if the font won't work on SDF. So it relies on _atleast_ always have 1 default font renderer. With this change that's no longer a possibility.

Since we can no longer can rely on CanvasTextRenderer always being there as a default, it means that `Stage.resolveTextRenderer()` can either return a renderer or `null` because:
* There are no text renderers at all or;
* The text renderer configured can't handle your font (i.e. SdfTextRenderer and you're trying to use a web font family).

Checking if there are 0 TextRenderers configured on Stage.createTextNode is easy and solves the you have no text rendering configured at all. But in the latter case above, you have SDF configured but are trying to create a node with a unsupported font family, the textRenderer can still resolve to `null`. Since the Canvas Text Renderer might not be there. Throw an error at that point stops the assignment but the Node is already created and the Renderer will find it in the tree and do frame update processing.

We could move the checks to stop `update()` and `checkRenderProps()` on the CoreTextNode, but IMHO that would make it worse over checks on getters/setters. Even though these would just be 2 IF conditions, they just get called a lot more frequently over the getters/setters which typically get called once, at most (most get set on the constructor) or in some cases a little more frequent but never per frame (which update() does).

Another path we could move the Stage `resolveTextRenderer` prior to creating the CoreTextNode, that way we can completely halt the creation of the node if the textRenderer can't be resolved. Versus doing this in the CoreTextNode constructor as it is now. Which performance wise shouldn't matter since we're calling `resolveTextRenderer` regardless of the flow, just higher in the chain so we can stop creation of the node entirely if we can't resolve a text renderder.

Wdyt @elsassph ?

src/core/Stage.ts
elsassph: Strictly speaking, one could now provide other implementations now, isn't it?

src/core/CoreTextNode.ts
elsassph: I think it's fair if `CoreTextNode` throws if it doesn't resolve a renderer. You don't have to avoid it at `Stage.createTextNode` level

wouterlucas: Well you'd want to avoid creating the node and assigning to the tree all together, right?

src/core/Stage.ts
wouterlucas: Fair enough - though it is a little counter open source to encourage that, I'd rather have a new renderer (i.e. WebGPU) in here so everyone can benefit.

I guess this could help on PlayStation with ION - though we can handle that separately.

src/core/CoreTextNode.ts
elsassph: But it's an error situation - it's not meant to happen.

src/core/Stage.ts
elsassph: Precisely, if you provide a WebGPU renderer, you won't need to update this `renderEngine` type, as it'll extend `CoreRenderer`!

elsassph:
```typescript
interface IRendererConstructor {
  new(options: CoreRendererOptions): CoreRenderer;
}

export interface StageOptions {
  ...
  renderer: IRendererConstructor;
}
```

jfboeve (approved):

src/core/Stage.ts
wouterlucas: Will look into dealing with opening up engine types later, moving this along

erikhaandrikman (approved):

PR #355, opened 2024-08-13
fix: add get / set for data object

chiefcll:
The inspector allows setting a custom data object. However some also use it for `data` from API. Adding the get / set to allow devs to access the data that they set.


wouterlucas (approved):
LGTM

Issue #349, opened 2024-07-25
Ability to prevent a node from being GC'ed


wouterlucas:
Add a flag on a CoreNode/CoreTextNode that would prevent it from being cleaned up by the memory management routine, regardless if that is over its limits or not.

This helps for textures that are offscreen but guaranteed to be on screen frequently or shortly, the App Dev framework can indicate what nodes should never be cleaned up regardless of the memory pressure limits.

PR #356, opened 2024-08-13
 Ability to prevent a node from being GC'ed #349

Drulokia:


src/core/TextureMemoryManager.ts
```diff
@@ -227,8 +227,10 @@ export class TextureMemoryManager {
         // We don't want to free renderable textures because they will just likely be reloaded in the next frame
         break;
       }
-      texture.ctxTexture.free();
-      txManager.removeTextureFromCache(texture);
+      if (!texture.preventCleanup) {
```
wouterlucas, line 230: `texture.preventCleanup === false` will perform slightly better over a fallsy check, since we know its always going to be a boolean, might as well.

wouterlucas (changes requested):

src/core/TextureMemoryManager.ts
elsassph: Hard to believe, and I'm 99% sure that minification process will remove this "optimisation".

chiefcll:
Are there details on use case for this? Or some docs to reference? Is this primarily for Images rather than simply nodes?

wouterlucas:
> Are there details on use case for this? Or some docs to reference? Is this primarily for Images rather than simply nodes?

Exactly that, to prevent textures on a node from being cleaned up even when the memory threshold has been exceeded and/or they are out of the view port. It helps when you have nodes with textures that you know you're going to need nearly immediate;y regardless of whether it is in view port or not.

wouterlucas (approved):
LGTM

PR #361, opened 2024-08-21
Fix: Add support for mount property in absX and absY

### LINKED ISSUE #354: Properties absX and absY do not consider mount value
Hello, if I'm not mistaken properties `absX` and `absY` in `CoreNode` should be the real position of the component in the page, calculated based on the X and Y values of the component itself and all its parents. The value should also consider other properties like `mount`, `scale`, `pivot` and `rotation` to calculate the exact position.

I did some test and this code works by handling the `mount` property;

```js
// src/core/CoreNode.ts row 1401
	get absX(): number {
		return (
			this.props.x -
			this.props.width * this.props.mountX +
			(this.props.parent?.absX || this.props.parent?.globalTransform?.tx || 0)
		);
	}

	get absY(): number {
		return this.props.y - this.props.height * this.props.mountY + (this.props.parent?.absY ?? 0);
	}
```

I'm trying to support mouse/magic remote, so the real position of the component is really important.

scailbc:
Fixes #354:

- include `mount` property when calculating `absX` and `absY`
- create new example page `absolute-position` to test `absX` and `absY` with different values of `mount`, `scale`, `rotation` and `pivot`

### DIFF
--- a/examples/tests/absolute-position.ts
+++ b/examples/tests/absolute-position.ts
@@ -1,0 +1,146 @@
+import type { INode } from '@lightningjs/renderer';
+import type { ExampleSettings } from '../common/ExampleSettings.js';
+
+export default async function ({ renderer, testRoot }: ExampleSettings) {
+  function createTextNode(container: INode, label?: string | undefined) {
+    const textNode = renderer.createTextNode({
+      x: 10,
+      y: 10,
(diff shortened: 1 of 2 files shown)


wouterlucas (approved):
LGTM - thanks!

PR #362, opened 2024-08-22
perf: cleanup loop - extra ? checks

chiefcll:


src/core/Stage.ts
```diff
@@ -387,18 +387,13 @@ export class Stage {
       node.renderQuads(this.renderer);
     }
 
-    for (let i = 0; i < node.children.length; i++) {
-      const child = node.children[i];
+    const childrenLength = node.children.length;
+    for (let i = 0; i < childrenLength; i++) {
+      const child = node.children[i] as CoreNode;
 
-      if (!child) {
```
elsassph, line 393: I think `if (child)` isn't a luxury

chiefcll: That was only added for typescript. You shouldn't have a child which doesn't exist

src/core/Stage.ts
```diff
@@ -387,18 +387,13 @@ export class Stage {
...
-        continue;
+      if (child.worldAlpha !== 0) {
+        this.addQuads(child);
```
wouterlucas, line 395: I'd prefer we keep it `=== 0` as a "off case" and keep the normal path down below as it was, just reads better in my opinion.

src/core/Stage.ts
wouterlucas: it can if the array was modified while you started the loop?

Issue #352, opened 2024-08-02
Alpha wrong on nodes with Radius Shader


chiefcll:
I have a Row of images which change alpha from 0.6 to 1.0 based on being focused. When the images are Rounded via Radius Shader, the alpha is wrong for the next items. It seems to be correct for the first two items in the row but as I keep switching to the next item it breaks. 

<img width="1043" alt="image" src="https://github.com/user-attachments/assets/2476ad39-d1b2-47b0-b71b-b495ffbba78c">

This is the same row of images without the Radius Shader on the thumbnails and it works fine. You'll notice even the focused Alpha doesn't go back up to 1. 
<img width="1028" alt="image" src="https://github.com/user-attachments/assets/6aa50686-51e7-431d-8ec6-57de3b2d7377">

Drulokia:
I cannot reproduce this, do you have a code example of this?

wouterlucas:
With @Drulokia being on holiday - yes having a starting point would help.

PR #366, opened 2024-09-03
fix alpha issue dynamic shader

jfboeve:
Resolved alpha issue with shaders #352. Traced it back to automatic alpha value not being updated on time, and it not being properly tested when checking if it can be batch rendered by the dynamic shader.


wouterlucas (approved):
Sweet! LGTM

PR #367, opened 2024-09-03
updated typedoc config

jfboeve:
I noticed the current typedoc config was not up to date since the split from threadx, and the canvas/webgl changes. This fix exposes all the good stuff again.


wouterlucas (approved):
LGTM

Issue #368, opened 2024-09-03
ImageWorker not working on WPE 2.22


chiefcll:
If you have time to try ImageWorker on WPE2.22 isn't working for me. I haven't had time to debug yet. Current work around is just setting numImageWorkers=0

PR #369, opened 2024-09-04
fix: Safari 11 doesn't have ImageBitmap

chiefcll:


wouterlucas:
Is this related to https://github.com/lightning-js/renderer/issues/368 ?

Because AFAIK `createImageBitmap` is suppose to work on WPE 2.22 - unless its configured incorrectly.

src/core/renderers/canvas/CanvasCoreTexture.ts
```diff
@@ -131,7 +131,7 @@ export class CanvasCoreTexture extends CoreContextTexture {
       if (ctx) ctx.putImageData(data, 0, 0);
       this.image = canvas;
       return { width: data.width, height: data.height };
-    } else if (data instanceof ImageBitmap) {
+    } else if (window.ImageBitmap && data instanceof ImageBitmap) {
```
wouterlucas, line 134: I'm a little confused, how is a check `window.ImageBitmap` required of you're checking `data instanceof ImageBitmap`. 

Does this check no longer work in this particular case:
https://github.com/lightning-js/renderer/blob/main/src/core/textures/ImageTexture.ts#L145

or more specifically is this check failing? 
`!!self.createImageBitmap;`
https://github.com/lightning-js/renderer/blob/main/src/core/CoreTextureManager.ts#L158

We need to catch it earlier up the chain if `createImageBitmap` doesn't work or if we're incorrectly detecting its presence. 

Also `window.ImageBitmap` will break future multithreading webworker support, we should not introduce window specific code knowing we want to run this in a webworker in the future (and thus kicking the can down the road).

src/core/renderers/webgl/WebGlCoreCtxTexture.ts
```diff
@@ -134,7 +134,7 @@ export class WebGlCoreCtxTexture extends CoreContextTexture {
     // If textureData is null, the texture is empty (0, 0) and we don't need to
     // upload any data to the GPU.
     if (
-      textureData.data instanceof ImageBitmap ||
+      (window.ImageBitmap && textureData.data instanceof ImageBitmap) ||
```
wouterlucas, line 137: Same as above. This won't work in a worker - as there is no `window` object and it's best to catch this earlier up the chain if `imageBitmap` doesn't work.

wouterlucas (changes requested):

src/core/renderers/canvas/CanvasCoreTexture.ts
elsassph: You need this check, but it could be written as 
```
if (typeof ImageBitmap !== 'undefined' && data instanceof ImageBitmap)
```

src/core/renderers/webgl/WebGlCoreCtxTexture.ts
elsassph: How can you check earlier up the chain? This code needs to test all the possible inputs it can receive.

wouterlucas: if `createImageBitmap` isn't available, you shouldn't be using `ImageBitmaps` at all - that way this can't happen?

chiefcll:
https://github.com/rdkcentral/Lightning/blob/240563c426d6669f9aedcca28524b35bca3f5a7c/src/platforms/browser/WebPlatform.mjs#L139

src/core/renderers/webgl/WebGlCoreCtxTexture.ts
chiefcll: PS - this isn't running in a worker.

wouterlucas: ah I see - the `ImageBitmap` object doesn't exists at all - therefor the IF will always throw an error. Having more context + a stack trace and a somewhat decent error description helps, would appreciate if you can be a little bit descriptive next time. Thanks for adding context @elsassph that OP forgot to add.

Agree, in that case `typeof ImageBitmap !== 'undefined'` is safest and avoids future issues with `window` references.

wouterlucas (approved):
LGTM

PR #381, opened 2024-09-13
Remove uniform wrapper and use concise setUniform<1f,2f,3f,4f> methods

### LINKED ISSUE #374: Unnecessary heap allocations per frame for WebGL renderer.
https://github.com/rdkcentral/Lightning/issues/494

Reopening this issue which impacted Lightning 2 and still impacts Lightning 3.

https://github.com/lightning-js/renderer/blob/3a1fb19183a7b77914e31fff62b187bcf1484e7e/src/core/renderers/webgl/WebGlCoreShader.ts#L313-L320

https://github.com/lightning-js/renderer/blob/3a1fb19183a7b77914e31fff62b187bcf1484e7e/src/core/renderers/webgl/WebGlCoreShader.ts#L313-L320

The direct usage of the generic version of this function leads to each of those spread and rest operators creating new arrays that have to be garbage collected after each frame that includes these shaders.

If `...args` was replaced with `arguments` usage, it is possible you could keep the generic function without the explicit array being created from the rest operator usage.

For general applications this is fine, but the moment you have an animation that includes a re-render this can quickly grow into forcing many garbage collection cycles.

wouterlucas:
Fixes #374 

Could not fall back to `arguments` as that isn't allowed in strict mode, working around that only made it worse.
Moved everything up early in the chain to call the respective uniform method directly, with minimal overhead which should avoid the additional GC pressure as described in #374  

Only thing I need to figure out what the impact is of losing the `setUniform` argument caching that was originally there, worst case we have to implement a cache check at each individual uniform method.

@jfboeve the Dynamic Shader was a bit of a thing - please 👀 if this makes sense. I don't know if the additional argument uniforms can actually happen in the Dynamic Shader case (I assumed so).


src/core/renderers/webgl/shaders/DynamicShader.ts
```diff
@@ -162,7 +162,45 @@ export class DynamicShader extends WebGlCoreShader {
...
           uniform?.updateProgramValue!(effect.props[key], props);
         }
-        this.setUniform(uniformInfo[key]!.name, effect.props[key].programValue);
+
+        const uniform = uniformInfo[key]!.uniform;
+        const value = effect.props[key].programValue;
+        const location = this.getUniformLocation(uniformInfo[key]!.name);
+
+        // check single argument uniforms first
+        if (
```
elsassph, line 171: if `props.effects` doesn't change, could you resolve the right functions once at init time so here you know what to call already?

elsassph: Or maybe check whether `value` is an array and what length it is

wouterlucas: Yeah - though `value` can also be a Float32Array where it needs to be provided as-is to the GL uniform. Can also do a bunch of type checks and then handle that accordingly, which will likely be a little shorter then the current expanded version.

What I do like about this is that if you provide it a wrong `value`, you'll get the appropriate `uniform` error directly from GL without us being in the middle. Since we're just mapping uniform type to the function underneath it and calling it as is.

jfboeve (approved):

PR #383, opened 2024-09-16
perf: Replace forEach with for loops

chiefcll:
Hopefully reduces a bit of garbage collection. Only removing forEach on most used areas of the code. Also updated eslint to disable the no null assertion so typescript doesn't complain about undefined children.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -40,7 +40,7 @@ const config = {
     '@typescript-eslint/no-unsafe-assignment': 'warn',
     '@typescript-eslint/no-unsafe-return': 'warn',
     '@typescript-eslint/no-unsafe-call': 'warn',
-    '@typescript-eslint/no-non-null-assertion': 'off',
+    '@typescript-eslint/no-non-null-assertion': 'warn',
     '@typescript-eslint/no-unsafe-member-access': 'warn',
     '@typescript-eslint/restrict-template-expressions': [
       'warn',
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -40,6 +40,7 @@ const config = {
     '@typescript-eslint/no-unsafe-assignment': 'warn',
     '@typescript-eslint/no-unsafe-return': 'warn',
     '@typescript-eslint/no-unsafe-call': 'warn',
+    '@typescript-eslint/no-non-null-assertion': 'off',
     '@typescript-eslint/no-unsafe-member-access': 'warn',
     '@typescript-eslint/restrict-template-expressions': [
       'warn',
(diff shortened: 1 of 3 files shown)


src/core/CoreNode.ts
```diff
@@ -976,9 +976,9 @@ export class CoreNode extends EventEmitter {
       this.updateType ^ UpdateType.All &&
       this.updateType & UpdateType.RenderTexture
     ) {
-      this.children.forEach((child) => {
-        child.setUpdateType(UpdateType.All);
-      });
+      for (let i = 0, length = this.children.length; i < length; i++) {
```
wouterlucas, line 979: is there a reason to do an `length = this.children.length` over `i < this.children.length`?
Wouldn't `length` also need a `let` or `const` declaration?

If this is the same as taking `const length = this.children.length` prior to the for loop, which I believe Jeffrey has done on a couple of places before. I remember there was no measurable different between the two patterns.

src/core/CoreNode.ts
```diff
@@ -976,9 +976,9 @@ export class CoreNode extends EventEmitter {
...
+        this.children[i]!.setUpdateType(UpdateType.All);
```
wouterlucas, line 980: You have to be careful with the TS `!` postfix, it does two things: checks for `undefined` and `null` whereas you can get away with just a `undefined` check here. Plus it gets transpiled to exactly that anyway.

.eslintrc.cjs
```diff
@@ -40,6 +40,7 @@ const config = {
     '@typescript-eslint/no-unsafe-assignment': 'warn',
     '@typescript-eslint/no-unsafe-return': 'warn',
     '@typescript-eslint/no-unsafe-call': 'warn',
+    '@typescript-eslint/no-non-null-assertion': 'off',
```
wouterlucas, line 43: I don't think we'd want to turn this off - a warning would be fine.

wouterlucas (changes requested):

src/core/CoreNode.ts
chiefcll: The ! tells typescript to ignore the possible error of an undefined coming back from the array due to accessing out of index - https://www.executeprogram.com/courses/typescript-basics/lessons/undefined-in-arrays#:~:text=There's%20a%20case%20where%20TypeScript,case%2C%20TypeScript%20will%20return%20undefined%20.

Doing a `if (this.children !== undefined)` is an extra check and doesn't get compiled away.

src/core/CoreNode.ts
chiefcll: I'm kinda surprised as you'd be checking the length of the array each time during the loop. It should be faster to store the length as the array length could change and must be rechecked each loop.

https://stackoverflow.com/questions/8452317/do-loops-check-the-array-length-every-time-when-comparing-i-against-array-length

wouterlucas: yeah that used to be an issue - seems to be less a problem with newer webkits/chromium last time we spend a morning playing with WPEWebKit and different chromium versions in . Eitherway its all very marginal and doesn't hurt to keep it that way.

wouterlucas (approved):
LGTM

PR #385, opened 2024-09-17
compat: Remove fromEntries for older browser support

chiefcll:
Functionally the same thing - fromEntries requires Chrome 73. This will allow renderer to work on older devices a bit easier without vite needing to transpile it away.

### DIFF
--- a/src/core/text-rendering/renderers/TextRenderer.ts
+++ b/src/core/text-rendering/renderers/TextRenderer.ts
@@ -431,23 +431,24 @@ export abstract class TextRenderer<
     };
     // For each prop setter add a wrapper method that checks if the prop is
     // different before calling the setter
-    this.set = Object.freeze(
-      Object.entries(propSetters).reduce((acc, [key, setter]) => {
-        acc[key as keyof TrProps] = (
-          state: StateT,
-          value: TrProps[keyof TrProps],
-        ) => {
+    const propSet = {};
+    Object.keys(propSetters).forEach((key) => {
+      Object.defineProperty(propSet, key, {
+        value: (state: StateT, value: TrProps[keyof TrProps]) => {
+          // Check if the current prop value is different before calling the setter
           if (state.props[key as keyof TrProps] !== value) {
-            setter(state, value as never);
+            propSetters[key as keyof TrPropSetters](state, value as never);
+
             // Assume any prop change will require a render
-            // This is required because otherwise a paused RAF will result
-            // in renders when text props are changed.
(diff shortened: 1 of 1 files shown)


src/core/text-rendering/renderers/TextRenderer.ts
```diff
@@ -432,23 +432,22 @@ export abstract class TextRenderer<
     // For each prop setter add a wrapper method that checks if the prop is
     // different before calling the setter
     this.set = Object.freeze(
```
elsassph, line 434: While we look at that: `Object.freeze` is notoriously slow in older browsers.

src/core/text-rendering/renderers/TextRenderer.ts
```diff
@@ -432,23 +432,22 @@ export abstract class TextRenderer<
...
+            setter(state, value as never);
+            // Assume any prop change will require a render
+            // This is required because otherwise a paused RAF will result
+            // in renders when text props are changed.
+            this.stage.requestRender();
+          }
+        };
+        return acc;
+      }, {} as TrPropSetters<StateT>),
+    );
```
wouterlucas, line 450: why not avoid entries and reduce all together with a simple for loop?
Entries would still be polyfilled anyway since its Chrome 54 I believe? Save some more and go all the way.

Something like:
```js
const propSettersMap: TrPropSetters<StateT> = {};

const keys = Object.keys(propSetters);
for (let i = 0; i < keys.length; i++) {
  const key = keys[i] as keyof TrProps;
  const setter = propSetters[key];

  propSettersMap[key] = (state: StateT, value: TrProps[key]) => {
    if (state.props[key] !== value) {
      setter(state, value as never);
      this.stage.requestRender();
    }
  };
}
```

should do it?

src/core/text-rendering/renderers/TextRenderer.ts
wouterlucas: Agree - though you can't get similar immutability that freeze gives you. Object.assign comes close enough for the usecase.

Something like:
```js
this.set = {};
Object.keys(propSetters).forEach(key => {
  Object.defineProperty(this.set, key, {
    value: propSetters[key],
    writable: false,  // Prevents property from being changed
    configurable: false  // Prevents property from being deleted
  });
});
```

And replace that with the `propSetterMap` and we should be good to go?

wouterlucas (changes requested):

wouterlucas (approved):
LGTM

PR #390, opened 2024-09-20
fix: re-evaluate isRenderable on alpha change

### LINKED ISSUE #372: Nodes with alpha = 0 can't have their src changed
I have two background images which switch src and fade in / out from alpha 0 to 1. When the nodes go to alpha 0, the changes to node is ignored, even if I alpha is changed back to 1.

To reproduce:

create a node with a src
set node to alpha = 0
change the src
set node to alpha = 1

Node will not come back

### LINKED ISSUE #388: Textures being freed?
This is running on my macbook - sometimes the FPS update counter text just goes away. Also the logos and text at the top left also is gone. This happens regularly on TV's but first time I've seen it on my macbook as well.

Screenshot 2024-09-18 at 6 05 37 PM

wouterlucas:
On low powered devices / heavy operation UI's there might be a race condition where the animation/transition is ends prior to the update loop finishing its first run. This causes the isRenderable to not be evaluated in time for the alpha changes to take affect.

This PR ensures isRenderable is always re-evaluated on an alpha change and moves the worldAlpha calculation slightly up to ensure it gets handled in the correct order if its a 1 shot update. 

Also save a little processing by ensuring on alpha change the children only re-evaluate the worldAlpha and not a global, reducing child update processing.

Fixes #388 
Might fix #372  - though cant reproduce to validate on my setup (even with 20x slowdown).

### DIFF
--- a/src/core/CoreNode.ts
+++ b/src/core/CoreNode.ts
@@ -1027,6 +1027,20 @@ export class CoreNode extends EventEmitter {
       }
     }
 
+    if (this.updateType & UpdateType.WorldAlpha) {
+      if (parent) {
+        this.worldAlpha = parent.worldAlpha * this.props.alpha;
+      } else {
+        this.worldAlpha = this.props.alpha;
+      }
+      this.setUpdateType(
+        UpdateType.Children |
+          UpdateType.PremultipliedColors |
+          UpdateType.IsRenderable,
+      );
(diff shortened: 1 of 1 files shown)


erikhaandrikman (approved):
LGTM!

Drulokia:
This indeed seems to fix the issue in #372

PR #392, opened 2024-09-23
Fix toggle clipping & boundaries

wouterlucas:
Because we where using the boundaries from parent as an `out` parameter, the new variable for `this.parent.strictBoundary` became a reference not a memcopy. This is fine most of the time but if the user would repeatedly turn clipping on / off the `strictBoundary` wouldn't reset to the appropriate parent value. 

This PR changes to make sure the `strictBoundaries` are new variable copies versus references.


visual-regression/certified-snapshots/chromium-ci/viewport-events-11.png
jfboeve, line 1: There is something wrong here, the Blue rect is off screen but the Status says its in bounds

wouterlucas: because its in that 100 px in-bound boundary, 12 shows it completely out of bounds

erikhaandrikman (approved):
LGTM

PR #394, opened 2024-09-25
Flag parent if render texture child has changes

erikhaandrikman:
Fixes an issue where out-of-bounds or non-renderable child nodes of a render texture remain invisible


wouterlucas (approved):
LGTM

PR #400, opened 2024-09-27
removed useProgram call WebGlCoreShader

jfboeve:
useProgram was used two times, so I removed one line.


erikhaandrikman (approved):
Nice catch :)

PR #401, opened 2024-09-27
Scheduled state for animations with delay

michielvandergeest:
Added `scheduled` state for animations with delay. `scheduled` will change to `running` as soon as the  animation _really_ starts (triggered by the `onAnimating` hook).

This allows to make a distinction between an animation that is actually in the process of animating (i.e. `running`) and animations that are awaiting the delay to start animating (i.e. `scheduled`).


wouterlucas (approved):
LGTM

erikhaandrikman (approved):

PR #404, opened 2024-10-01
Introduce a strictBounds flag

wouterlucas:
Enabling strictBounds will not render / process nodes that are outside of the visible area, otherwise the default behavior will render all nodes regardless of their boundary status.


erikhaandrikman (approved):
LGTM

elsassph:
Why is that even an option?

wouterlucas:
> Why is that even an option?

It was causing too much havoc - I don't think all clients are ready for super strict boundary rendering yet.
We can always flip the default down the line.

Alternatively we can do a check to see if the node has children that are outside of their parent boundaries and implement behavior on that, but this comes at extra costs.

For now, we can set it to true by default where we need it to be and toggle this from an app dev perspective.

PR #406, opened 2024-10-04
Fail gracefully createNativeCtxTexture in WebGlCoreCtxTexture load

CosmaVergari:
- Add try/catch block in WebGlCoreCtxTexture `load` method to manage the error thrown by `createNativeCtxTexture`
- Set the texture state to failed on error thrown

### DIFF
--- a/src/core/renderers/webgl/WebGlCoreCtxTexture.ts
+++ b/src/core/renderers/webgl/WebGlCoreCtxTexture.ts
@@ -85,12 +85,12 @@ export class WebGlCoreCtxTexture extends CoreContextTexture {
     }
     this._state = 'loading';
     this.textureSource.setState('loading');
-    try {
-      this._nativeCtxTexture = this.createNativeCtxTexture();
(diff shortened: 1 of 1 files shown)


wouterlucas:
Try / Catch statements are notoriously dangerous for their performance impact, as essentially you're copying the stack just in case _something_ might happen. 

What is the error exactly and isn't this handled in the `catch` already?

CosmaVergari:
I just made a change to address your point. The idea is to manage the error raised by `createNativeCtxTexture` [here](https://github.com/lightning-js/renderer/blob/dd9461f55ae7a2946141af934c84d88335c937bb/src/core/renderers/webgl/WebGlCoreCtxTexture.ts#L248)

CosmaVergari:
Thank you for the suggestion, I converted the raised error to a `null` return. The performance gain seems significant indeed

wouterlucas:
Thanks for your contribution!

wouterlucas (approved):

PR #408, opened 2024-10-07
Refactor example font URLs in installFonts.ts and update base path in Fonts

wouterlucas:
Required to fix the auto deploy to gh pages


michielvandergeest (approved):
LGTM

PR #409, opened 2024-10-08
Performance updates

wouterlucas:
Stash of small performance updates:
* Removal of lots of redundant `const` creations
* Introduced a Texture Type to avoid more expensive `instanceof` checks
* Small refactor on `flipY` detection with RTT
* Reduced immediate GC pressure on node destroy by setting values to `undefined` over explicit `delete`
* Skip sanitizing custom `data` on node creation if inspector is disabled

Performance changes:
```
create: 60.24ms ±17.30 ->  50.07ms ±8.84
update: 30.66ms ±1.50 -> 28.87ms ±9.51
skipNth: 24.32ms ±7.00  -> 22.81ms ±5.22
select: 27.72ms ±12.20  -> 26.44ms ±15.72
swap: 28.14ms ±14.80 -> 24.16ms ±12.71
remove: 25.72ms ±5.20 -> 22.83ms ±5.25
createLots: 303.16ms ±141.40  -> 258.30ms ±50.86
append: 282.90ms ±36.70 -> 283.57ms ±191.68
clear: 14.76ms ±1.50 -> 14.59ms ±5.86
```


src/core/CoreNode.ts
```diff
@@ -1459,24 +1460,23 @@ export class CoreNode extends EventEmitter {
...
 
     this.props.texture = null;
     this.props.shader = this.stage.defShaderCtr;
 
-    const children = [...this.children];
-    for (let i = 0, length = children.length; i < length; i++) {
-      children[i]!.destroy();
+    for (let i = 0, length = this.children.length; i < length; i++) {
+      this.children[i]?.destroy();
     }
```
chiefcll, line 1478: The length of children was changing so it wasn't deleting every node. I'd recommend double checking this doesn't introduce a memory leak. Or you just always delete this.children[0] until empty.

wouterlucas: OK agree, doing a while loop until empty is then a better solution. Spreading a new array into a const is just slightly more expensive and creates additional GC pressure that in turn slows the whole thing down again.

wouterlucas: while loop should address it! lmk if it doesnt and thanks for the call out.

jfboeve (approved):

PR #410, opened 2024-10-08
fix: fallback when fetch does not supports URL scheme "data"

pecoram:
This PR implements a fallback for cases where createImageBitmap is supported but the fetch fails because does not supports URL scheme "data" (for example when using a polyfill).

### DIFF
--- a/src/core/lib/utils.ts
+++ b/src/core/lib/utils.ts
@@ -285,3 +285,42 @@ export function convertUrlToAbsolute(url: string): string {
   const absoluteUrl = new URL(url, self.location.href);
   return absoluteUrl.href;
 }
+
+export function base64DataFromSrc(src: string): [string, string] {
+  const matches = src.match(/^data:(.*?);base64,(.*)$/);
+
+  let mimeType;
+  let base64Content;
+
+  if (matches) {
+    mimeType = matches[1];
+    base64Content = matches[2];
+  }
+
+  mimeType = mimeType ?? 'image/png';
+  base64Content = base64Content ?? src;
+  base64Content = base64Content.replace(/-/g, '+').replace(/_/g, '/');
+
+  return [base64Content, mimeType];
+}
+
+export function base64toBlob(base64Data: string, contentType: string) {
+  contentType = contentType || '';
+  const sliceSize = 1024;
+  const byteCharacters = atob(base64Data);
+  const bytesLength = byteCharacters.length;
+  const slicesCount = Math.ceil(bytesLength / sliceSize);
+  const byteArrays = new Array(slicesCount);
+
+  for (let sliceIndex = 0; sliceIndex < slicesCount; ++sliceIndex) {
+    const begin = sliceIndex * sliceSize;
+    const end = Math.min(begin + sliceSize, bytesLength);
+
+    const bytes = new Array(end - begin);
+    for (let offset = begin, i = 0; offset < end; ++i, ++offset) {
+      bytes[i] = byteCharacters[offset]?.charCodeAt(0);
+    }
+    byteArrays[sliceIndex] = new Uint8Array(bytes);
+  }
+  return new Blob(byteArrays, { type: contentType });
+}
--- a/src/core/lib/utils.ts
+++ b/src/core/lib/utils.ts
@@ -286,7 +286,7 @@ export function convertUrlToAbsolute(url: string): string {
   return absoluteUrl.href;
 }
 
-export function base64DataFromSrc(src: string): [string, string] {
+export function base64DataFromSrc(src: string): [string, string] | undefined {
   const matches = src.match(/^data:(.*?);base64,(.*)$/);
 
   let mimeType;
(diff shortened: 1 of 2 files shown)


src/core/textures/ImageTexture.ts
```diff
@@ -143,8 +147,15 @@ export class ImageTexture extends Texture {
...
       );
     } else if (this.txManager.hasCreateImageBitmap === true) {
-      const response = await fetch(src);
-      const blob = await response.blob();
+      let blob;
+      try {
+        const response = await fetch(src);
+        blob = await response.blob();
+      } catch (error) {
+        const base64data = base64DataFromSrc(src);
```
elsassph, line 155: This looks very unsafe:

- There can be many errors caught here,
- It seems to be blindly trying to parse `src` as a data URL, while this seem like a very specific case.

pecoram: like this is better?

src/core/textures/ImageTexture.ts
```diff
@@ -143,8 +147,19 @@ export class ImageTexture extends Texture {
         sh,
       );
     } else if (this.txManager.hasCreateImageBitmap === true) {
-      const response = await fetch(src);
-      const blob = await response.blob();
+      let blob;
+      try {
```
wouterlucas, line 151: if `fetch` is a promise, why wrap it in a try/catch block?

src/core/textures/ImageTexture.ts
```diff
@@ -143,8 +147,19 @@ export class ImageTexture extends Texture {
...
+        const response = await fetch(src);
+        blob = await response.blob();
+      } catch (error) {
+        if (src.startsWith('data:')) {
```
wouterlucas, line 155: wouldn't it make more sense to do the `src.startsWith('data:')` check up top before or just after we do `fetch`? Sounds expensive to always make the fetch fail first, deal with the error chain and then try to parse a `data` blob. 

If its only an error case you're worried about, why not: 
```ts
const response = await fetch(src).catch( e => { /* .... handle error */ });

if (!response && src.startWith('data:')) {
  /* ... do alternate parsing here */
}
```

If this if this is a 1 in 10 or 1 in 100 time case, going through the try/catch error chain to then figure out its a `data:` and do an alternative path is very expensive.

src/core/lib/utils.ts
```diff
@@ -285,3 +285,42 @@ export function convertUrlToAbsolute(url: string): string {
...
+  let mimeType;
+  let base64Content;
+
+  if (matches) {
+    mimeType = matches[1];
+    base64Content = matches[2];
+  }
+
+  mimeType = mimeType ?? 'image/png';
+  base64Content = base64Content ?? src;
```
wouterlucas, line 301: if it didn't match, it would still return src and then fail down below in the base64toBlob?
I think we should handle the error case more strictly if we can figure it is not Base64 and bail out gracefully.

src/core/textures/ImageTexture.ts
wouterlucas: I honestly think this is a bad idea, rather keep things straight and neat instead of trying to fallback and make it work. If the feature is not available, we shouldn't try to make it work. It will cause confusion.

wouterlucas (changes requested):
Honestly I don't think we should do fallback when we can predetermine that the `createImageBitmap` won't work.

In such case we should just rely on `Image()` and turn off the path that involves `createImageBitmap`.

src/core/textures/ImageTexture.ts
pecoram: What can I do is using a function like this:

```
async function supportsDataURLFetch() {
  try {
    await fetch('data:text/plain,Test');
    return true;
  } catch (error) {
    return false;
  }
}
```

and check it at startup

src/core/textures/ImageTexture.ts
elsassph: It's idiomatic nowadays to `try { await } catch`

wouterlucas: is it? that doesn't automatically mean its a good idea 🙃

src/core/textures/ImageTexture.ts
wouterlucas: two questions:
* Does polyfilling it actually helps over just turning createImageBitmap support off?
* What polyfill are you using, as this seems like limitations of the polyfill?

pecoram: I used the polyfill (https://gist.github.com/pseudosavant/a6d970b945ae85ef4dbc43200da94faf) because the images were not taking the correct size, but with polyfill it was giving me that exception. Then I realised that by removing width and height from the new Image everything worked correctly and then I removed the polyfill. I added the fallback in the PR to avoid possible problems like this.

pecoram:
OK - I'm opening the PR with just the change we need, then we'll consider whether to go ahead with this

PR #412, opened 2024-10-14
fix: image size on browsers without createImageBitmap

pecoram:
This PR addresses an issue specific to 2017 Samsung TVs, which do not support the createImageBitmap method.

**Testing**

Tested on 2017 Samsung TVs with renderer+solid.


wouterlucas (approved):
LGTM

PR #418, opened 2024-10-21
Resolved sans-serif font issue for firefox browser

jfboeve:
The FireFox browser was throwing an error when using the default font for canvas. 

When the canvas text renderer tries to add the default font (fallback font).  However during the creation of the WebFont (line: 120 CanvasTextRenderer) it creates a FontFace which creates a family property which is an empty string in FireFox. This is where we run into problems because that empty string is used as a key later on. 

I have resolved the issue by building in a fallback that in case the family returns an empty string it falls back to the fontFamily property of its super class.

### DIFF
--- a/src/core/text-rendering/renderers/CanvasTextRenderer.ts
+++ b/src/core/text-rendering/renderers/CanvasTextRenderer.ts
@@ -242,28 +242,22 @@ export class CanvasTextRenderer extends TextRenderer<CanvasTextRendererState> {
     // the `isFontFaceSupported` check)
     assertTruthy(fontFace instanceof WebTrFontFace);
 
+    const fontFamily = fontFace.fontFamily;
+
     // Add the font face to the document
     // Except for the 'sans-serif' font family, which the Renderer provides
     // as a special default fallback.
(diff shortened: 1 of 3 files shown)


src/core/text-rendering/renderers/CanvasTextRenderer.ts
```diff
@@ -252,7 +252,11 @@ export class CanvasTextRenderer extends TextRenderer<CanvasTextRendererState> {
     }
 
     const { fontFamilies } = this;
-    const familyName = fontFace.fontFace.family;
+
+    let familyName = fontFace.fontFace.family;
+    if (familyName === '') {
+      familyName = fontFace.fontFamily;
```
wouterlucas, line 258: is it worth to check if `fontFace.fontFamily` exists? or is this a last resort anyway?

wouterlucas (approved):

src/core/text-rendering/renderers/CanvasTextRenderer.ts
jfboeve: good point cleaned up the function

wouterlucas (approved):

PR #421, opened 2024-10-29
Set gl.clearColor to transparent for render-to-texture nodes

erikhaandrikman:
**Issue**: When rendering parts on the render tree to a texture ( RTT ) and the `clearColor` is set to a solid color rather than being transparent, this color is used to clear the colorbuffer of the texture's framebuffer. This results in render textures displaying a solid background color instead of being transparent.

This PR addresses this issues by setting the `gl.clearColor` to transparent before the render to texture pass begins and restoring it when the pass is complete.

Before: 

before

After:

after


wouterlucas (approved):
LGTM

PR #425, opened 2024-10-31
fix: added support to border and radius for canvas2d

pecoram:


src/core/renderers/canvas/CanvasCoreRenderer.ts
```diff
@@ -134,8 +135,22 @@ export class CanvasCoreRenderer extends CoreRenderer {
     const hasClipping = clippingRect.width !== 0 && clippingRect.height !== 0;
     const hasGradient = colorTl !== colorTr || colorTl !== colorBr;
     const radius = quad.shader ? getRadius(quad) : 0;
+    const border = quad.shader ? getBorder(quad) : undefined;
+    const borderTop = quad.shader ? getBorder(quad, 'Top') : undefined;
```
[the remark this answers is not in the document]
pecoram: I am undecided whether it is better to give freedom to override one of the borders (eg. border+borderBottom) or make them exclusive

pecoram:
there is a problem with the position of the border... on the webgl renderer the border is inside (which I think is wrong or at most to be made configurable) while on the canvas renderer it is centered

src/core/renderers/canvas/CanvasCoreRenderer.ts
```diff
@@ -134,8 +135,26 @@ export class CanvasCoreRenderer extends CoreRenderer {
     const hasClipping = clippingRect.width !== 0 && clippingRect.height !== 0;
     const hasGradient = colorTl !== colorTr || colorTl !== colorBr;
     const radius = quad.shader ? getRadius(quad) : 0;
+    const border = quad.shader ? getBorder(quad) : undefined;
+    const borderTop =
+      !border?.width && quad.shader ? getBorder(quad, 'Top') : undefined;
```
elsassph, line 140: You could avoid this complex test (esp when transpiled to ES5) by having `getBorder` returning `undefined` when there is no width.

elsassph: Also maybe:
```typescript
    let tryBorderShader = Boolean(quad.shader);
    const border = tryBorderShader ? getBorder(quad) : undefined; // assuming `undefined` if no `border.width`
    if (boder) tryBorderShader = false;
    const borderTop = tryBorderShader ? getBorder(quad, 'Top') : undefined;
    //etc
```

src/core/renderers/canvas/CanvasCoreRenderer.ts
```diff
@@ -134,8 +135,26 @@ export class CanvasCoreRenderer extends CoreRenderer {
...
+      !border?.width && quad.shader ? getBorder(quad, 'Bottom') : undefined;
+    const borderLeft =
+      !border?.width && quad.shader ? getBorder(quad, 'Left') : undefined;
 
-    if (hasTransform || hasClipping || radius) {
+    if (
+      hasTransform ||
+      hasClipping ||
+      radius ||
+      border ||
```
elsassph, line 152: Do we need to save the context for borders?

wouterlucas:
Nice, thanks for the contribution! I like how you've isolated it in the Canvas implementation and thanks for the extended review @elsassph!

Heads up @jfboeve has been working on a redesign on the shaders/effects portion of the renderer. Stepping away from the Dynamic Shader and having tree shakeable shaders.

One of the topics we discussed is shaders and Canvas2D and we're keeping those separate. Limiting Effects to WebGL1/WebGL2 only as it does do not make sense to combine. Trying to come up with something to caters to both WebGL & Canvas2D would only lead to overly complicated effects definition for only a small number of applicable effects for Canvas2D anyway.

We'll have to refactor this PR a little bit to make it fit with the new effects/shaders structure down the road.

wouterlucas (approved):

src/core/renderers/canvas/CanvasCoreRenderer.ts
```diff
@@ -211,6 +214,94 @@ export class CanvasCoreRenderer extends CoreRenderer {
...
+      } else {
+        ctx.strokeRect(
+          tx + borderInnerWidth,
+          ty + borderInnerWidth,
+          width - borderWidth,
+          height - borderWidth,
+        );
+      }
+      ctx.globalAlpha = 1;
+    } else {
```
elsassph, line 244: `else if (hasQuadShader) {` at least

elsassph (approved):

src/core/renderers/canvas/shaders/UnsupportedShader.ts
```diff
@@ -27,9 +27,9 @@ export class UnsupportedShader extends CoreShader {
   constructor(shType: string) {
     super();
     this.shType = shType;
-    if (shType !== ROUNDED_RECTANGLE_SHADER_TYPE) {
-      console.warn('Unsupported shader:', shType);
-    }
+    // if (shType !== ROUNDED_RECTANGLE_SHADER_TYPE) {
```
elsassph, line 30: Should that be removed?

pecoram: let's decide, having added the missing “shaders” I think it is no longer necessary. In addition, this warning makes the inspector unusable

PR #432, opened 2024-11-04
fix: [Inspector] drop Proxies, set selective Object.defineProperties traps

### LINKED ISSUE #429: Inspector RTT support is broken
Inspector seems to break when RTT nodes are enabled, it even stalls the rendering itself.

Without inspector:
image (2)

With inspector enabled:
image (3)

With inspector enabled and RTT disabled:
image (4)

wouterlucas:
Proxies introduce a delay and differences in update graph behaviour which break the RTT parent/child updates on nodes itself, to avoid the issue only enable traps for inspector elements that we're actually interested in.

Fixes #429

### DIFF
--- a/src/main-api/Inspector.ts
+++ b/src/main-api/Inspector.ts
@@ -270,13 +270,6 @@ export class Inspector {
     node: CoreNode | CoreTextNode,
     div: HTMLElement,
   ): CoreNode | CoreTextNode {
-    const updateNodePropertyWrapper = (
-      property: keyof CoreNodeProps,
-      value: any,
-    ) => {
-      queueMicrotask(() => this.updateNodeProperty(div, property, value));
-    };
-
     // Define traps for each property in knownProperties
     knownProperties.forEach((property) => {
       const originalProp = Object.getOwnPropertyDescriptor(node, property);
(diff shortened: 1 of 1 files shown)


src/main-api/Inspector.ts
```diff
@@ -239,72 +247,87 @@ export class Inspector {
...
-      },
-      get: (target, property: keyof CoreNode, receiver: any): any => {
-        if (property === 'destroy') {
-          this.destroyNode(target.id);
-        }
+    const updateNodePropertyWrapper = (
+      property: keyof CoreNodeProps,
+      value: any,
+    ) => {
+      queueMicrotask(() => this.updateNodeProperty(div, property, value));
```
elsassph, line 277: Is the delay useful?

wouterlucas: The issue wasn't on the performance of the DOM nodes but the impact Proxies and the inspector updates had on the CoreNode processing itself.

Though I don't mind taking it out if we see risk/concern on the DOM side, can always add it back if we need it.

elsassph: JS is synchronous so you will only slow down things a bit.

michielvandergeest (approved):
LGTM

Issue #436, opened 2024-11-06
Inspector breaks animationController


chiefcll:
```js
return {
          ...animationController,
          start: () => {
            this.animateNode(div, props, settings);
            return animationController.start();
          },
        };
 ```

Spreading the animationController doesn't work - its missing all the functions from EventEmitter base class. I think you need to do another defineProperty wrap around the controller for start method.

PR #437, opened 2024-11-06
fix: Inspector breaks animationController

pecoram:
this fixes https://github.com/lightning-js/renderer/issues/436


wouterlucas (approved):
Nice - thanks!

PR #451, opened 2024-11-20
Added method to set clearColor during runtime

Drulokia:
Provide a way to change the clearColor during runtime


jfboeve (changes requested):
Hey! The implementation looks good, however it would be nice if you could change the following regarding the test;
- `dynamic-settings.ts` to `clear-color-setting.ts`
- follow the following template to add it to the virtual regression test [here](https://github.com/lightning-js/renderer/blob/main/examples/tests/shader-animation.ts)

PR #453, opened 2024-11-21
fix: Fix RTT Clipping and improve RTT parent lookup

### LINKED ISSUE #443: Cache Parent RTT node for faster update behaviour
Optimization leftover from #441

### LINKED ISSUE #452: RTT & Clipping is not working
The moment you turn on clipping on a `rtt: true` node the node disappears. 

This is because the new update logic from parent -> children provides the original clipping rectangle, however since the children are part of render to texture the clipping rectangle does not apply as its clipped already as part of the RTT creation.

In the previous implementation children of RTT where given an invalid clipping rectangle to avoid incorrect clipping in Render texture coordinates.

wouterlucas:
Fixes #452 
Fixes #443

### DIFF
--- a/examples/tests/rtt-dimension.ts
+++ b/examples/tests/rtt-dimension.ts
@@ -263,6 +263,7 @@ export default async function test({ renderer, testRoot }: ExampleSettings) {
     switch (i) {
       case 1:
         rttNode.rtt = false;
+        rttNode.clipping = false;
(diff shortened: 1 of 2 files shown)


jfboeve (commented):
It looks like test rtt-dimension-2 is failing... does the test itself need updating? or did something not work as expected?

jfboeve (approved):

PR #456, opened 2024-11-25
Fixes for Chrome v38

wouterlucas:
Fixes for handling base64 src strings on Chrome v38, rework Image texture fallback & detection.

This PR includes:
* Using a 1x1 PNG pixel to determine which `createImageBitmap` signature is supported on the browser
* Fallback the image worker and main thread `ImageTexture` handling according to the detected signature
* Support for base64 `src` on Chrome v38
* Support for `createImageBitmap(blob)` without a config object for Chrome v50 (linked to #454)
* Support for ES5 legacy output of the `examples` directory (will be published on https://lightning-js.github.io/renderer/ automatically)
* Upgrade `vite` and `vitest` to 2.x devDependencies, pin to TS 5.6.x for now (as 5.7.x has issues with other dependencies)
* Upgrade Playwright 1.39 to 1.49 for automated tests
* Refactor Visual Regression test pixel matching and snapshot handling by using [pixelmatch](https://www.npmjs.com/package/pixelmatch) and [pngjs](https://www.npmjs.com/package/pngjs) over a homebrew detection + upng (which is stale).
* SDF font initialization waits for image support detection before uploading the SDF texture atlas.
* Added documentation on the fallback process and references to chrome versions in `BROWSERS.md`


src/core/textures/ImageTexture.ts
```diff
@@ -132,61 +132,79 @@ export class ImageTexture extends Texture {
     return mimeType.indexOf('image/png') !== -1;
   }
 
-  async loadImage(src: string) {
-    const { premultiplyAlpha, sx, sy, sw, sh, width, height } = this.props;
+  async loadImageFallback(src: string, hasAlpha: boolean) {
+    const img = new Image();
+    img.src = src;
```
elsassph, line 137: L2 sets the `src` after the event handlers.

src/core/CoreTextureManager.ts
```diff
@@ -183,11 +183,25 @@ export class CoreTextureManager {
...
+        this.hasCreateImageBitmap = result;
+
+        if (!this.hasCreateImageBitmap) {
+          console.warn(
+            '[Lightning] createImageBitmap is not supported on this browser. ImageTexture will be slower.',
+          );
+        }
+
+        if (this.hasWorker && numImageWorkers > 0) {
+          this.imageWorkerManager = new ImageWorkerManager(numImageWorkers);
```
elsassph, line 197: Assigned already line 183

wouterlucas: ah, yeah the one on line 183 needs to be removed. There is no point in spawning ImageWorkers if there is no createImageBitmap support. Thanks for the spot!

wouterlucas:
Using a 1x1 binary png pixel to detect if which createImageBitmap to handle chrome v50/51, small detour with eslint and dependencies to upgrade to newer vite legacy for es5 examples.

src/core/CoreTextureManager.ts
```diff
@@ -156,6 +162,12 @@ export class CoreTextureManager {
 
   imageWorkerManager: ImageWorkerManager | null = null;
   hasCreateImageBitmap = !!self.createImageBitmap;
+  imageBitmapSupported = {
+    basic: false,
+    options: false,
+    full: false,
+  };
```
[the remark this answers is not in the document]
wouterlucas: While I agree that enums can improve code clarity, in this specific case it introduces a limitation: it assumes that support levels are mutually exclusive.t doing that.

Our current approach uses two boolean flags (basic and advanced) to represent whether:
- The environment supports basic createImageBitmap(blob) functionality.
- The environment supports advanced options like createImageBitmap(blob, config).

This allows us to express combinations such as:
- basic: true, config: false, full false (Supports basic but not advanced options)
- basic: true, config: false, full: true (Doesn't support config alone, but does support cropping + config)
- basic: false, config: true, full: true (Full support)

If we want to stick to an enum we could use bitwise flags for more flexibility but IMHO that diminishes the readability and as such I decided to stick to simple booleans.

I don't know if there is a usecase for mixed support of features on `createImageBitmap`, but at the rate we're exploring these really outlandish browsers I chose to opt for a more flexible solution that could deal with this quite easily.

jfboeve (approved):

PR #458, opened 2024-11-27
Correct clipping of children in render-to-texture nodes

erikhaandrikman:
This PR resolves an issue where children nodes rendered as part of a render-to-texture were not properly clipped according to the parent node's bounds. The fix ensures that clipping is correctly applied within the render-to-texture pipeline, maintaining visual consistency and expected behavior for nested scenes


wouterlucas (approved):
LGTM

jfboeve (approved):

PR #462, opened 2024-12-02
fix: data attr in inspector not reactive..

chiefcll:
Data attr missing in list for Proxy so any changes to data was not reflected in inspector output.


wouterlucas:
Build is iffy because of race conditions with createImageBitmap on the test runner.

wouterlucas (approved):
LGTM

PR #467, opened 2024-12-10
428 dynamic bounds margin

Drulokia:
Added option to set boundsmargin on node level.
Also able to change it after creation.


jfboeve (changes requested):
Hey I went through the code the initial work looks good, however I've found some points I would like to see added / changed:

For the Stage.ts:
- Set the default boundsMargin of the stage when creating the rootNode.
- Add getter/setter for boundsMargin on Stage and have it also adjust the boundsMargin of the rootNode.

For the CoreNode:
- for `get boundsMargin` I don't think we need to check if the boundsMargin in the props is an Array or not, just check if this.props.boundsMargin is not null and return that value. If it is null let it check the parent node(s) with the changes to Stage we won't need the stage.boundsMargin
- In case the boundsMargin for a CoreNode is set upon creating the node we should call the boundsMargin setter like we are doing with the shaders/texture (line: 761)

wouterlucas (approved):
LGTM

Drulokia:
Added dynamic boundsmargin option to stage, also boundsmargin can be set to null, so it can take over parents boundsmargin or stages boundsmargin.

Went throught the changes requested by @jfboeve

jfboeve:
I agree with you @wouterlucas since the renderBounds are calculated based on boundsMargin it's better to add it to the RenderBound

Drulokia:
I had it seperate because the renderBounds did only propagate 1 child up.
I checked if it had impact if it did all children and it doesn't.
Changed it to be handled in renderBounds UpdateType

wouterlucas (changes requested):

wouterlucas (approved):
Thanks!

jfboeve (approved):

wouterlucas:
can you rebase it @Drulokia ?

wouterlucas:
I think you had a merge issue with the unit tests, seems broken after the rebase/merge with main:
https://github.com/lightning-js/renderer/actions/runs/12904794534/job/35982692173?pr=467#step:8:30

PR #469, opened 2024-12-10
fix: Origin error when using Image() constructor

wouterlucas:
This fixes the error:
net::ERR_BLOCKED_BY_RESPONSE.NotSameOriginAfterDefaultedToSameOriginByCoep 302 (Found)

### DIFF
--- a/src/core/textures/ImageTexture.ts
+++ b/src/core/textures/ImageTexture.ts
@@ -134,6 +134,7 @@ export class ImageTexture extends Texture {
 
   async loadImageFallback(src: string, hasAlpha: boolean) {
     const img = new Image();
+    img.crossOrigin = 'anonymous';
 
     return new Promise<{ data: HTMLImageElement; premultiplyAlpha: boolean }>(
       (resolve) => {


src/core/textures/ImageTexture.ts
```diff
@@ -134,6 +134,7 @@ export class ImageTexture extends Texture {
 
   async loadImageFallback(src: string, hasAlpha: boolean) {
     const img = new Image();
+    img.crossOrigin = 'anonymous';
```
elsassph, line 137: I think Lightning 2 do it as well, but has some custom logic do NOT do it for Playstation.
Maybe don't autodetect the platform, but `ImageTexture` could have a static flag which could be externally disabled to not do set this crossOrigin.

pecoram, line 137: Before it was excluded for inline images:
```suggestion
        if (!(src.startsWith('data:'))) {
      img.crossOrigin = 'Anonymous';
    }
```

pecoram (changes requested):

src/core/textures/ImageTexture.ts
wouterlucas: Funny enough that is only on SVG:
https://github.com/rdkcentral/Lightning/blob/fa1f699e8b00f087923875e72990bafc73893237/src/tools/Tools.mjs#L179

wouterlucas: yes! good catch.

wouterlucas: PS4 is currently set to go the ION route if all goes well - so skipping this until it pops up as an issue again.

PR #471, opened 2024-12-11
feat: add optional external canvas for rendering

wouterlucas:
This is used to already create a canvas object and render a logo, pass this on to Lightning and it will be overwritten once the app comes up


jfboeve (approved):
LGTM

PR #472, opened 2024-12-11
Introduce Texture Throttling

wouterlucas:
Batch the amount of textures being downloaded (source) and uploaded (core ctx creation) to the GPU to reduce load on constrained devices.

The default is set to `0` where everything will be processed in a single frame, this can be reduced to any value (e.g. `50` or `25`) to limit the amount of textures being processed per frame.

Todo:
- [x] Fix loaded events w/ dimensions
- [x] Run through all the tests (+fixes)
- [x] Check Canvas2D regression
- [x] Align RGB / RGBA handling with hasAlpha on Texture Creates to save some bits


src/core/renderers/CoreContextTexture.ts
jfboeve, line 1: I wonder if it's worthwhile to make a TextureState enum so we won't have to compare strings?

jfboeve (approved):

PR #473, opened 2024-12-12
fix: safari 9 HTMLImageElement check

pecoram:


src/core/renderers/webgl/internal/RendererUtils.ts
```diff
@@ -141,9 +141,10 @@ export function createIndexBuffer(glw: WebGlContextWrapper, size: number) {
 export function isHTMLImageElement(obj: unknown): obj is HTMLImageElement {
   return (
     obj !== null &&
-    typeof obj === 'object' &&
-    obj.constructor &&
-    obj.constructor.name === 'HTMLImageElement'
+    ((typeof obj === 'object' &&
+      obj.constructor &&
+      obj.constructor.name === 'HTMLImageElement') ||
+      obj instanceof HTMLImageElement)
```
elsassph, line 147: would it be safe/better to do the `instanceof` check first?

src/core/renderers/webgl/internal/RendererUtils.ts
```diff
@@ -141,9 +141,11 @@ export function createIndexBuffer(glw: WebGlContextWrapper, size: number) {
 export function isHTMLImageElement(obj: unknown): obj is HTMLImageElement {
   return (
     obj !== null &&
-    typeof obj === 'object' &&
-    obj.constructor &&
-    obj.constructor.name === 'HTMLImageElement'
+    ((typeof obj === 'object' &&
+      obj.constructor &&
+      obj.constructor.name === 'HTMLImageElement') ||
+      (typeof HTMLImageElement !== 'undefined' &&
```
wouterlucas, line 147: Hopefully this works if we're rendering in a webworker again - though granted if we're doing offscreen rendering the chances are that we're using Canvas rendering are quite low. 

That plus ThreadX being months out from a useable state again, its fine.

src/core/textures/ImageTexture.ts
```diff
@@ -134,9 +134,6 @@ export class ImageTexture extends Texture {
 
   async loadImageFallback(src: string, hasAlpha: boolean) {
     const img = new Image();
-        if (!(src.startsWith('data:'))) {
-      img.crossOrigin = 'Anonymous';
```
wouterlucas, line 138: is this giving issues for you? Because I actually needed this on my end if I wanted to use the new Image() constructor over createImageBitmap.

pecoram: there was the same code twice, I removed one of them

wouterlucas (approved):

PR #476, opened 2025-01-07
🎨 Refactor CoreNode Renderability Logic 🚀

wouterlucas:
### What’s New? ✨
This PR refactors Texture Throttling in the following ways:

- 🕒 Async RTT Texture Loading: When an RTT node is set, wait for the texture to load asynchronously.
- 🗑️ Improved Texture Cleanup: When a texture is freed, its source texture is now also marked as freed. Fixes a bug where textures weren't returning after memory cleanup.
- 🔄 Auto-Trigger Texture Loading: Refactored isRenderable owner changes to automatically call texture loading if the texture state is freed or initial.

## How to Test? 🧪

RTT Changes: Use `test=rtt-dimension`. 🖼️ You should see race conditions occurring prior to this PR.

Texture Freed Fix: Add a rocko.png to the `test=texture-cleanup-critical` test and move it in and out of the screen to validate. 🎯 (Not the prettiest, but it works!). Might create an automated test for this in the future.

## What’s Next? 🚀
More Testing! 🛠️

### DIFF
--- a/src/core/CoreNode.ts
+++ b/src/core/CoreNode.ts
@@ -1017,7 +1017,7 @@ export class CoreNode extends EventEmitter {
     if (this.updateType & UpdateType.RenderTexture && this.rtt) {
       // Only the RTT node itself triggers `renderToTexture`
       this.hasRTTupdates = true;
-      this.stage.renderer?.renderToTexture(this);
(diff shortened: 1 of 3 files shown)


jfboeve (approved):
LGTM!

PR #479, opened 2025-01-09
Refactor isRenderable detection

### LINKED ISSUE #477: All nodes are renderable
I was looking to fix: https://github.com/lightning-js/renderer/issues/463 and noticed it wasn't happening any longer in main. This is because all nodes are now renderable due to 
https://github.com/lightning-js/renderer/blob/main/src/core/CoreNode.ts#L771

Every node is being given a texture making it renderable. I assume you don't want that to be the case as every container node is now being rendered.

For instance:
```js
const yellowRect = renderer.createNode({
    x: 300,
    y: 0,
    width: 100,
    height: 100,
    color: 0x00000000,
    parent: testRoot,
  });
```
Gets the defaultTexture, even though it shouldn't be rendered.

wouterlucas:
✨ What changed? ✨ 

* Decouple **isRenderable** from **hasRenderableOwner** as textures can be loaded asynchronously.
* Refactored the `isRenderable` detection:
  - If a node has `alpha = 0` or is off screen its not renderable
  - If a node has dimensions and color properties, its marked renderable and we assume its a `ColorTexture`.
  - If a node has a texture, it will be marked as a renderable owner. Which will trigger loading of the texture (if its the first owner)
    - If the texture is loaded, it is marked renderable
* On `renderQuads` we assume everything has a texture if not we set the defaultTexture for 1x1 sampling. This is implicit but matches the old behaviour.
* `CoreTextNode` overwrites behaviour and checks for text !== ''. 
* Also fixed a bug where the default texture was still loading but already referenced.
* Added a `CoreNode` unit test to validate this more rapidly. 

🔧 Addresses the following
Fixes #477

### DIFF
--- a/src/core/CoreNode.ts
+++ b/src/core/CoreNode.ts
@@ -768,7 +768,16 @@ export class CoreNode extends EventEmitter {
         UpdateType.RenderState,
     );
 
-    this.createDefaultTexture();
+    // if the default texture isn't loaded yet, wait for it to load
+    // this only happens when the node is created before the stage is ready
+    if (
+      this.stage.defaultTexture &&
+      this.stage.defaultTexture.state !== 'loaded'
+    ) {
+      this.stage.defaultTexture.once('loaded', () => {
+        this.setUpdateType(UpdateType.IsRenderable);
+      });
+    }
   }
 
   //#region Textures
@@ -806,18 +815,6 @@ export class CoreNode extends EventEmitter {
         this.onTextureFreed(texture);
       }
     });
-  }
-
-  createDefaultTexture(): void {
-    // load default texture if no texture is set
-    if (
-      this.stage.defaultTexture !== null &&
-      this.props.src === null &&
-      this.props.texture === null &&
-      this.props.rtt === false
-    ) {
-      this.texture = this.stage.defaultTexture;
-    }
   }
 
   unloadTexture(): void {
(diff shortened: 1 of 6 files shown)


src/core/CoreNode.ts
```diff
@@ -768,7 +768,16 @@ export class CoreNode extends EventEmitter {
...
+    // if the default texture isn't loaded yet, wait for it to load
+    // this only happens when the node is created before the stage is ready
+    if (
+      this.stage.defaultTexture &&
+      this.stage.defaultTexture.state !== 'loaded'
+    ) {
+      this.stage.defaultTexture.once('loaded', () => {
+        this.setUpdateType(UpdateType.IsRenderable);
+      });
+    }
```
chiefcll, line 780: Would it be better to force defaultTexture to be loaded before a node can be created?

src/core/CoreNode.ts
```diff
@@ -1391,28 +1348,100 @@ export class CoreNode extends EventEmitter {
...
-      this.isRenderable = newIsRenderable;
-      this.onChangeIsRenderable(newIsRenderable);
+  /**
+   * Sets the renderable state and triggers changes if necessary.
+   * @param isRenderable - The new renderable state
+   */
+  setRenderable(isRenderable: boolean) {
+    if (this.isRenderable !== isRenderable) {
+      this.isRenderable = isRenderable;
     }
```
chiefcll, line 1404: don't think you need the if anymore

src/core/CoreNode.ts
```diff
@@ -1563,7 +1591,9 @@ export class CoreNode extends EventEmitter {
       colorTr: this.premultipliedColorTr,
       colorBl: this.premultipliedColorBl,
       colorBr: this.premultipliedColorBr,
-      texture: this.texture,
+      // if we do not have a texture, use the default texture
+      // this assumes any renderable node is either a distinct texture or a ColorTexture
+      texture: this.texture || this.stage.defaultTexture,
```
chiefcll, line 1596: Why not default this in the constructor?

src/core/CoreNode.ts
wouterlucas: no that means we'd have to hold off on node creation until the Stage is fully initialized, requiring queue's or events to hold off a platform to start creating nodes.

That's not worth the trouble if we can just wait for textures to be loaded asynchronously.

src/core/CoreNode.ts
wouterlucas: Fair - this was a left over pattern.

src/core/CoreNode.ts
wouterlucas: That's how it was before the change. That would assign a default Texture to everything again which complicated the detection if we're renderable, which is what we don't want. Or we have to detect what kind of node you're trying to create and re-evaluate that on every change for various properties (src, rtt, texture assignment, etc). The default texture is only required to sample Color rectangles off of and only make sense if we have a w/h and a color (i.e. `ColorTexture`).

That means we'll have to check if we have color properties + dimensions on the constructor to detect if its a `ColorTexture`. However if you then set the `src` after creation or something like that we have to have reevaluate if we're still a `ColorTexture` or e.g. an `ImageTexture` now.

I had that initially however that lead to a pattern to do redetection / rechecking if we require a `ColorTexture` or not, leading to rechecking the same thing for a lot of values changed on various properties (RTT, src, Texture, etc).

The pattern we have now is quite similar to the old pattern prior to TextureThrottling. It would go into the `renderQuads` and at CoreCtx creation figure out if we have an assigned Texture Type if not apply the default. In that sense this follows a similar approach. 

Since it can't be anything _but_ a `ColorTexture` at this point because:
* We wouldn't be renderable, thus wouldn't be in `renderQuads` if we're not renderable
* If it already had a texture it was anything else then a `ColorTexture`
* Thus assume its a `ColorTexture` if texture isn't set.

jfboeve (approved):

PR #480, opened 2025-01-10
fix: added missing check for shader existence

pecoram:
this fix the issue that elements with shaders but without color are not marked as renderable


wouterlucas (approved):
Sweet! Thanks :)

cc @erikhaandrikman - yeah we where wrong about the shader assumption 😬

PR #494, opened 2025-01-23
Fix #455 incorrect width when using width and contain

### LINKED ISSUE #455: Width of text node with word wrapping not correct when text is smaller than word wrapping width
There seems to be a bug (or at least an unexpected result from a developer perspective), when creating a text node with "word wrapping" applied to it (i.e. a `width` property and `contain` property set to `both`).

When the text content is _larger_ than the width property, everything works as expected. But whenever the text content is _smaller_ than the desired word wrapping width, the final width of the texture reported in the text loaded event is not the actual text width, but equals the value of the supplied `width`).

From a developer perspective, one would expect that the `width` property is interpreted as a _max width_, and the returned `width` to be _actual_ width of the final texture.

Drulokia:


wouterlucas (approved):
LGTM

PR #497, opened 2025-01-23
Texture lifecycle improvements

wouterlucas:
This PR introduces several improvements and fixes related to the handling of textures in the rendering pipeline. 

* Addressed Texture Reloading Issues: Resolved inconsistencies that occurred when textures transitioned back to a renderable state after being previously marked non-renderable.
* Preserve Source Data: Adjusted the cleanup process to ensure only the GPU texture is cleared when a texture becomes non-renderable, retaining the source data for potential reuse.
* Propagate SubTexture Events: Added functionality to forward SubTexture freed events if the parent texture undergoes cleanup, ensuring proper event handling and lifecycle alignment.
* Introduced comprehensive tests to validate the entire texture lifecycle, ensuring robustness and preventing regressions in future updates. (See screenshot below for test validation details.)

texture-reload-1


src/core/textures/Texture.ts
```diff
@@ -262,10 +272,18 @@ export abstract class Texture extends EventEmitter {
    */
   free(): void {
     this.ctxTexture?.free();
-    if (this.textureData !== null) {
```
elsassph, line 265: That may be problematic for instance an app with a lot of rails. You want the images to be unloaded.
Maybe and `ImageTexture` with a `src` should still unload, as it can reload pretty quickly from network cache.

wouterlucas: Yea we have been debating that, we know that for WPEWebKit this is a fine approach as the NetworkProcess will already cache the images at the lower C++ levels. Newer Chrome's for sure as well.

However someone pointed out this might not apply for all browsers out there and we can't rely on that assumption. Fetching the source again might introduce too many delays if not cached by the browser. We can make it an config option, for now I've only made it an API on the Texture class.

It also brings up the issue that this is hooked to the Memory Management of the GPU while the data lives on the CPU HEAP. It feels functionally incorrect to clear HEAP data based on GPU metrics, unless we can come up with a more reliable metric to determine we need to clear anything. E.g. we could think about a separate image source cache with its own limits/control metrics.

I've moved it to a separate API call for now, but more then happy to rediscuss this. Current thinking is we keep the behavior the same as prior versions and L2. We can separately come up with a more refined approach to either control and/or manage image source caching.

wouterlucas: Moved convo to #499

PR #500, opened 2025-01-24
fix: Exclude Base64 images from image worker handling

pecoram:
In Chrome versions before 67, data: URLs were treated as uncertain-origin resources for security reasons. In this context:

XMLHttpRequest did not allow access to base64 (data:) URLs because it violated CORS (Cross-Origin Resource Sharing) policies. Even though a data: URL is "local," it was still considered a different origin from the domain where the request was made.

### DIFF
--- a/src/core/lib/ImageWorker.ts
+++ b/src/core/lib/ImageWorker.ts
@@ -68,71 +68,64 @@ function createImageWorker() {
       var supportsOptionsCreateImageBitmap =
         options.supportsOptionsCreateImageBitmap;
       var supportsFullCreateImageBitmap = options.supportsFullCreateImageBitmap;
-      var xhr = new XMLHttpRequest();
-      xhr.open('GET', src, true);
-      xhr.responseType = 'blob';
 
-      xhr.onload = function () {
-        if (xhr.status !== 200) {
-          return reject(new Error('Failed to load image: ' + xhr.statusText));
-        }
+      fetch(src)
+        .then((response) => response.blob())
+        .then((blob) => {
+          var withAlphaChannel =
+            premultiplyAlpha !== undefined
+              ? premultiplyAlpha
+              : hasAlphaChannel(blob.type);
 
-        var blob = xhr.response;
-        var withAlphaChannel =
-          premultiplyAlpha !== undefined
-            ? premultiplyAlpha
-            : hasAlphaChannel(blob.type);
-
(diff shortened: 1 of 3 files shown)


wouterlucas:
I think we should skip the image worker entirely if we know its a `data:` url - no point to push this over to a child process boundary just to decode source?

pecoram:
I agree! What do others think?

wouterlucas:
There's already a fetch function if you skip the image worker, it will automatically fallback to fetch which should work out of the box:
https://github.com/lightning-js/renderer/blob/main/src/core/textures/ImageTexture.ts#L208-L223

as far as I can see extending that if condition with a startsWith should do it?

src/core/lib/ImageWorker.ts
```diff
@@ -68,71 +68,64 @@ function createImageWorker() {
       var supportsOptionsCreateImageBitmap =
         options.supportsOptionsCreateImageBitmap;
       var supportsFullCreateImageBitmap = options.supportsFullCreateImageBitmap;
-      var xhr = new XMLHttpRequest();
```
wouterlucas, line 71: guess we can revert these changes now? as we're handling it on the main thread.

src/core/textures/ImageTexture.ts
```diff
@@ -135,7 +135,7 @@ export class ImageTexture extends Texture {
   async loadImageFallback(src: string, hasAlpha: boolean) {
     const img = new Image();
 
-    if (!src.startsWith('data:')) {
+    if (!isBase64Image(src)) {
```
wouterlucas, line 138: `=== false` please

src/core/textures/ImageTexture.ts
```diff
@@ -206,6 +206,7 @@ export class ImageTexture extends Texture {
 
     if (this.txManager.hasCreateImageBitmap === true) {
       if (
+        !isBase64Image(src) &&
```
wouterlucas, line 209: `=== false` please

wouterlucas (changes requested):

src/core/lib/ImageWorker.ts
pecoram: I think that fetch is faster and more secure than XMLHttpRequest and most browsers can benefit. But no problem. I will revert it

wouterlucas (approved):
LGTM! Thanks

PR #501, opened 2025-01-24
fix: subTexture in canvas renderer

pecoram:


src/core/renderers/canvas/CanvasCoreRenderer.ts
```diff
@@ -140,7 +141,10 @@ export class CanvasCoreRenderer extends CoreRenderer {
       if (texture.state === 'freed') {
         return;
       }
-      if (texture.state !== 'loaded') {
+      if (
+        texture.state !== 'loaded' ||
+        (textureType === TextureType.subTexture && !ctxTexture.hasImage())
```
wouterlucas, line 146: is this required? are there cases where the state is loaded but there is no image?

pecoram: @elsassph You added it in an earlier commit. Do you think this is necessary?

wouterlucas: AFAIK this was required pre texture throttling, with TT in place nodes that are not renderable wont get here and if you do not have an image or other type of texture you can't be rendered. The state handling / marking a node ready for rendering has moved up the chain.

elsassph: I think it was solving the problem that subtexture would not inherit the loaded state of the parent texture.

wouterlucas:
they do, and works on WebGL but since we moved to more refined texture types the Canvas Renderer was missing support for a SubTexture specifically.

PR #514, opened 2025-02-06
fix: resolve issues with fetching local assets when serving files from the file protocol

vonvick:
#### What does this PR do?
This PR fixes the issue where the assets in locally packaged apps are unresolved and fail to load. Now the local assets are loaded successfully and served to the app.

#### How has this been tested?
This has been tested locally on various WebOS devices

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,4 +1,4 @@
-lockfileVersion: '9.0'
+lockfileVersion: '6.0'
 
 settings:
   autoInstallPeers: true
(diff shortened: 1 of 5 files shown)


wouterlucas:
Thanks! I guess this would not work with Image workers set to `0` right?

Because then it'll fall back to:
https://github.com/lightning-js/renderer/blob/main/src/core/textures/ImageTexture.ts#L223

vonvick:
I need to test that with that setting. I am guessing that it would fail because it would be using fetch which does not support the `file` protocol. Do you reckon we can use the `XMLHttpRequest` there instead of the fetch?

> Thanks! I guess this would not work with Image workers set to `0` right?
> 
> Because then it'll fall back to: https://github.com/lightning-js/renderer/blob/main/src/core/textures/ImageTexture.ts#L223

michielvandergeest:
@vonvick yes we should use XHR there as well, to ensure we can load local files on the web worker as well. To my understanding there are no performance implications between XHR and fetch when loading simple images. For (large) JSON files, fetch supposedly is faster

Also we need to add license headers for the new helper file you created (`src/core/text-rendering/font-face-types/utils.ts`)

Other than that, this looks good to me!

vonvick:
@wouterlucas it did fail with the ImageWorker set to 0. But I have fixed it and now it works with or without the image worker being used.

cc @michielvandergeest

wouterlucas:
we should really rename `fetchJson` to be something more generic then that, since there is no `json` parsing anymore and its being used for images as well 😄

PR #517, opened 2025-02-10
fix programvalue update for length 4 arrays

jfboeve:
Fixed an issue where program values were not properly updated. The problem was in the Effect utils itself were simply the mapping was wrong.


michielvandergeest (approved):
nice find!

Issue #518, opened 2025-02-10
rtt memory allocation & lazy loading


wouterlucas:
Currently if you use RTT it loads everything regardless of its visibility state, meaning all RTT nodes are always loaded and it is counted double on the memory manager.

Ideally we want:
* RTT nodes to only load when they are inbound of their respective view port
* Memory counted should only account for the final texture generated, not the intermediate buffer used to generate the RTT node

Nodes loaded without RTT:

![Image](https://github.com/user-attachments/assets/fa31202c-4399-4a9b-94a4-2b12f68b9804)

Same test with RTT:

![Image](https://github.com/user-attachments/assets/a890d08c-f61a-4a71-bbd8-04b93ca73a3f)

From first glance at the Chrome task viewer it doesn't seem to increase the memory on the GPU, and that makes sense. Likely just double counted on the memory manager side. But TBD investigation on a device.

PR #519, opened 2025-02-10
Optimize Texture Throttling loading

wouterlucas:
This PR addresses the following issues:
* Textures being deleted while actively being in queue for Texture throttling --> Fixed by yanking them from the queue if orphaned
* Textures data being deleted while texture was being uploaded --> moved texture data deletion to the `coreCtx` handling so it is all in 1 place
* Textures being procssed for `download` while having no `download` aspect, this should mostly be for `ImageTextures` only --> `Color`, `Sub`, and `RTT` textures are directly pushed to the `upload` queue respectively
* Cleaned up excessive handling in the `subTexture` 

Added new test cases to reflect their real world-ish use cases. 

There is an ongoing issue with RTT and memory as outlined #518 - will push this out and investigate that separately.


src/core/CoreTextureManager.ts
```diff
@@ -437,10 +466,24 @@ export class CoreTextureManager extends EventEmitter {
...
     coreContext.load();
   }
 
+  /**
+   * Check if a texture is being processed
+   */
+  isProcessingTexture(texture: Texture): boolean {
+    return (
+      this.downloadTextureSourceQueue.includes(texture) ||
+      this.uploadTextureQueue.includes(texture)
```
michielvandergeest, line 483: this could be changed to a strict `true` check instead of a _truthy_ check

michielvandergeest (approved):
LGTM.

1 suggestion for a nano optimization 😄

jfboeve (approved):

PR #520, opened 2025-02-10
feat: expose cleanup feature

### LINKED ISSUE #516: add explicit Garbage Collect API
Add an API to expose a request to explicitly garbage collect textures (GFX memory).

Currently this is done at two levels:
* `onIdle` when over threshold but not critical
* per frame when `critical` threshold has been exceeded

However in some cases, e.g. heavy scene switches or with devices under severe memory constraints (looking at you Sky Q) we may want to cal this more often to avoid spilling over.

wouterlucas:
Fixes #516

### DIFF
--- a/src/core/Stage.ts
+++ b/src/core/Stage.ts
@@ -697,4 +697,14 @@ export class Stage {
       strictBounds: props.strictBounds ?? this.strictBounds,
     };
   }
+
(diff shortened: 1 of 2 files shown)


michielvandergeest (approved):
LGTM

PR #522, opened 2025-02-14
Fix text overflow calculation not correct

Drulokia:
It used to calculate width of text on last line always as needing the ellipsis even if that line would fit on the line.


Drulokia:
Adding Issue #523 and #268 to this pull request

Drulokia:
The wordBreak option in the layoutText function determines how words should be broken when they reach the end of a line. It is similar to the CSS word-break property. Here are the possible values and their effects:

1. normal: Words will break at normal word boundaries (e.g., spaces).
2. break-all: Words can break at any character, allowing for more aggressive line breaks.
3. break-word: Words will break only at word boundaries, but if a word is too long to fit on a line, 
it will be broken at the character level.

Default the renderer will use the normal wordBreak. So if you want long texts to have a overflow ellipses you have to change the wordBreak.

This is now simular how css does this: https://developer.mozilla.org/en-US/docs/Web/CSS/word-break
CSS-WordBreak

michielvandergeest (approved):
LGTM

@jfboeve would appreciate if you could also take an extra look at these changes.

jfboeve (approved):
Code looks good to me. We might wanna change xStart to startX in the future, same as the xOffset used in Glyphs but we can save that for the refactor I guess

yananazarenko417:
We also experience the same issue in our app. The text width is 253 px and the clip width is 270 - but the lightning renderer still truncates the title, although it fits perfectly fine. This breaks the marquee behaviour in our app
Could somebody re-check it please? I don't think it's been actually fixed
@Drulokia @wouterlucas @michielvandergeest @jfboeve

PR #526, opened 2025-02-19
GC enhancements

wouterlucas:
This change adds the following:
- An aggressive GC attempt if the quick GC fails to clean below critical threshold. Modelled after L2 like behaviour.
- Ability to block texture uploads if we're above critical memory. With a setting to turn that behaviour on or off: `doNotExceedCriticalThreshold`
   - This should help with devices that are constrained on GFX memory and do not recover easily when the GFX memory is maxed out
- A quick GC every idle loop, not only when going idle. Capped at 10ms
- Improved texture lifecycle to deal with textures that do not get uploaded because of memory constraints
- Fix memory alpha channel handling
- Introduce `createImageBitmapSupport` flag to be able to turn the auto detection on or off it allows for: `'auto' | 'basic' | 'options' | 'full'` - the original auto detect is still available under `auto`, default is set to `full`
   - This ensures a faster startup in cases where we know the createImageBitmap signature before hand


wouterlucas:
> Little comments, without having looked at the code:
> 
>     * this logic is broken in L2, the agressive GC never happens because of the security preventing GC to run more than once per frame,
> 
>     * it is very possible to go "over budget" in an app; when this happens, all the textures are actually required and doing a hard GC would just destroy the performance without fixing an actual problem.

The hard GC in L2 is also something you really don't want to happen 😆  it will also delete renderable textures. Which is what we're not doing in our case. The more thorough GC is just slightly more expensive and we'll have to iteratively feel how much of a hit on performance that is on the lower end devices, its just going through all textures sorted versus going through a more opportunistic list of orphaned textures in the quick GC.

One thing we could do is not sort it every time we need to do an aggressive GC - because that's likely the majority of the cycles spent without losing too much consistency. We can refine it if it turns out to be an issue.

On the second note, yeah _most_ devices are okay but we're currently targeting a device that does not like this at all. The screen will go black, it will flicker and never recover properly if you go over budget. The setting `doNotExceedCriticalThreshold` will just stop uploading new textures if we're over budget until more older textures are cleaned up. This should prevent it from going above that threshold all together, planning to enable that specifically for this device.

elsassph:
The problem is that when you go over budget, a GC won't solve it because the app will reload the textures anyway. It requires to rework the application so it doesn't happen.

src/core/Stage.ts
```diff
@@ -372,8 +377,13 @@ export class Stage {
...
     // Check if we need to cleanup textures
-    if (this.txMemManager.criticalCleanupRequested) {
-      this.txMemManager.cleanup();
+    if (this.txMemManager.criticalCleanupRequested === true) {
+      this.txMemManager.cleanup(false);
+
+      if (this.txMemManager.criticalCleanupRequested === true) {
+        // If we still need to cleanup, request another but aggressive cleanup
+        this.txMemManager.cleanup(true);
+      }
```
jfboeve, line 386: Is this correct?

jfboeve (approved):

PR #528, opened 2025-02-24
Feat/make renderer options usable at runtime

Drulokia:
Initial method for setting the renderer options runtime


jfboeve (approved):

michielvandergeest (approved):
LGTM

PR #537, opened 2025-03-06
Fix/rc3 optimisations

jfboeve:
- resolved issue where props were cloned obviously not needed to do that again.
- resolved issue where props where cloned into textprops.
- optimised check to see if the shader is updatable.
- optimised uniform passing where you can now pass an array and it will assign it for you without the need to create another float32array.

### DIFF
--- a/src/core/lib/ImageWorker.ts
+++ b/src/core/lib/ImageWorker.ts
@@ -158,7 +158,8 @@ function createImageWorker() {
       supportsFullCreateImageBitmap,
     })
       .then(function (data) {
-        self.postMessage({ id: id, src: src, data: data });
+        // @ts-ignore ts has wrong postMessage signature
+        self.postMessage({ id: id, src: src, data: data }, [data.data]);
       })
       .catch(function (error) {
         self.postMessage({ id: id, src: src, error: error.message });
--- a/src/core/lib/utils.ts
+++ b/src/core/lib/utils.ts
@@ -359,3 +359,28 @@ export function calcFactoredRadiusArray(
   result[3] *= factor;
   return result;
 }
+
+export function dataURIToBlob(dataURI: string): Blob {
+  dataURI = dataURI.replace(/^data:/, '');
+
+  const type = dataURI.match(/image\/[^;]+/)?.[0] || '';
+  const base64 = dataURI.replace(/^[^,]+,/, '');
+
+  const sliceSize = 1024;
+  const byteCharacters = atob(base64);
+  const bytesLength = byteCharacters.length;
+  const slicesCount = Math.ceil(bytesLength / sliceSize);
+  const byteArrays = new Array(slicesCount);
+
+  for (let sliceIndex = 0; sliceIndex < slicesCount; ++sliceIndex) {
+    const begin = sliceIndex * sliceSize;
+    const end = Math.min(begin + sliceSize, bytesLength);
+
+    const bytes = new Array(end - begin);
+    for (let offset = begin, i = 0; offset < end; ++i, ++offset) {
+      bytes[i] = byteCharacters[offset]?.charCodeAt(0);
+    }
+    byteArrays[sliceIndex] = new Uint8Array(bytes);
+  }
+  return new Blob(byteArrays, { type });
+}
--- a/src/core/textures/ImageTexture.ts
+++ b/src/core/textures/ImageTexture.ts
@@ -23,7 +23,11 @@ import {
   isCompressedTextureContainer,
   loadCompressedTexture,
 } from '../lib/textureCompression.js';
-import { convertUrlToAbsolute, isBase64Image } from '../lib/utils.js';
+import {
+  convertUrlToAbsolute,
+  dataURIToBlob,
+  isBase64Image,
+} from '../lib/utils.js';
 import { isSvgImage, loadSvg } from '../lib/textureSvg.js';
 import { fetchJson } from '../text-rendering/font-face-types/utils.js';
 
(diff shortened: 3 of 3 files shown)


src/core/CoreTextNode.ts
```diff
@@ -409,14 +409,14 @@ export class CoreTextNode extends CoreNode implements CoreTextNodeProps {
...
-
-    assertTruthy(this.globalTransform);
+    // if (this.parentHasRenderTexture && this.props.parent?.rtt) {
+    //   this.globalTransform = Matrix3d.identity();
+    //   if (this.localTransform) {
+    //     this.globalTransform.multiply(this.localTransform);
+    //   }
+    // }
+
+    // assertTruthy(this.globalTransform);
```
wouterlucas, line 419: does this break anything with RTT? or does the super RTT handling deal with it already?
If it doesn't break anything we should just clean it up

src/core/CoreNode.ts
```diff
@@ -765,15 +766,16 @@ export class CoreNode extends EventEmitter {
   constructor(readonly stage: Stage, props: CoreNodeProps) {
     super();
 
-    this.props = Object.assign({}, props, {
+    this.props = {
```
wouterlucas, line 769: moved away from the Object.assign without copy? was the spread faster?

src/core/Stage.ts
```diff
@@ -597,8 +597,7 @@ export class Stage {
 
   createTextNode(props: Partial<CoreTextNodeProps>) {
     const fontSize = props.fontSize ?? 16;
-    const resolvedProps = {
-      ...this.resolveNodeDefaults(props),
+    const resolvedProps = Object.assign(this.resolveNodeDefaults(props), {
```
wouterlucas, line 600: see prev comment, is object assign faster or slower then the spread?

src/core/renderers/webgl/WebGlRenderOp.ts
```diff
@@ -56,15 +58,32 @@ export class WebGlRenderOp extends CoreRenderOp {
...
   ) {
     super();
     this.buffers = quad.sdfBuffers || renderer.quadBufferCollection;
     this.shader = quad.shader as WebGlShaderNode;
+    this.width = quad.width;
+    this.height = quad.height;
+    this.clippingRect = quad.clippingRect;
+    this.parentHasRenderTexture = quad.parentHasRenderTexture;
+    this.framebufferDimensions = quad.framebufferDimensions;
+    this.rtt = quad.rtt;
```
wouterlucas, line 83: where is this `this.rtt` used now?

src/core/renderers/webgl/WebGlRenderer.ts
```diff
@@ -374,7 +372,7 @@ export class WebGlRenderer extends CoreRenderer {
     fQuadBuffer[bufferIdx++] = 1; //node y coord
 
     // Update the length of the current render op
-    this.curRenderOp.length += WORDS_PER_QUAD;
+    // this.curRenderOp.length += WORDS_PER_QUAD;
```
wouterlucas, line 375: ?

src/core/renderers/webgl/WebGlRenderOp.ts
wouterlucas: never mind found it down below :)

src/core/renderers/webgl/WebGlRenderer.ts
```diff
@@ -428,36 +426,41 @@ export class WebGlRenderer extends CoreRenderer {
...
       return false;
     }
 
     // Force new render operation if rendering to texture
     // @todo: This needs to be improved, render operations could also be reused
     // for rendering to texture
-    if (parentHasRenderTexture !== undefined || rtt !== undefined) {
+    if (
+      params.parentHasRenderTexture !== undefined ||
+      params.rtt !== undefined
```
wouterlucas, line 449: I guess here it is, instantiating a `const` over grabbing it from the object. Nice 👍

src/core/renderers/webgl/WebGlRenderer.ts
```diff
@@ -428,36 +426,41 @@ export class WebGlRenderer extends CoreRenderer {
...
+      return true;
+    }
+
     // Check if the shader can batch the shader properties
     if (
-      !this.curRenderOp.shader.program.reuseRenderOp(
-        params,
-        this.curRenderOp.quad as QuadOptions,
-      )
+      !this.curRenderOp.shader.program.reuseRenderOp(params, this.curRenderOp)
```
wouterlucas, line 463: `=== false`?

src/core/renderers/webgl/WebGlRenderer.ts
jfboeve: Was checking if this is doing anything for the code... turns out we don't do anything with the curRenderOp.length at all at the moment. So I think we can clean it up.

src/core/CoreNode.ts
jfboeve: The Object.assign without copy caused some issues with setting the parent, src, texture, shader below. So reverted this back to a spread.

src/core/Stage.ts
jfboeve: In comparison to the previous comment, there's no weird setup with setters we have to worry about so it's safe to use assign.

src/core/CoreTextNode.ts
jfboeve: This calculation is indeed already handled in the update function of the CoreNode (even in v2).

wouterlucas: alright then we should clean it up👍

wouterlucas (approved):
LGTM !

PR #539, opened 2025-03-07
feat: updated the way the compressed textures are detected

chrisjcnewton:
Instead of checking file extension for ktx and pvr image formats.
Load the file and check the file header to detect the format.
This makes it less brittle and allows urls without a file extension to work

### DIFF
--- a/src/core/lib/textureCompression.ts
+++ b/src/core/lib/textureCompression.ts
@@ -20,33 +20,76 @@
 import { type TextureData } from '../textures/Texture.js';
 
 /**
- * Tests if the given location is a compressed texture container
- * @param url
- * @remarks
- * This function is used to determine if the given image url is a compressed
- * and only supports the following extensions: .ktx and .pvr
- * @returns
+ * Checks if the texture at the given URL is in a compressed format (KTX or PVR).
+ *
+ * @param url - The URL of the texture to check.
+ * @returns A promise that resolves to the ArrayBuffer of the texture if it is in a compressed format, or null otherwise.
  */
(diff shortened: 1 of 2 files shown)


wouterlucas:
I think this is going to be quite expensive to do, basically if we can't determine it a provided `type` (which I don't think is supported by Blits atm @michielvandergeest ?) we will always try to:

- Fetch the image
- Check the buffers
- Determine if its compressed or not

While this may work well for actual compressed textures, this will absolutely kill regular `png` and `jpg` handling. Because if we've determined its not a compressed image, we're going to fetch it again and apply the regular image processing. 

I think you're going to destroy the performance of regular image loading with this.

Questions:
* Is it common for compressed textures to have an ambiguous extension?
* Could we provide a type from Blits @michielvandergeest to indicate its a compressed texture, avoiding the whole expensive processing step?

michielvandergeest:
This change was initially my suggestion to make the compressed texture detection a bit more sophisticated, rather then just looking at the file extension. I think we do the same for `.png` images, but I understand this adds execution time ..

I think the main point is that we can't 100% rely on file extensions, same as a png image can be hosted on a location without a file extension for example.

If we believe this is too impactful on performance (do we really fetch the file again if it's not a compressed texture file?), we can indeed think a solution where we pass a type from blits .. TDB how exactly, at the moment.

wouterlucas:
Closing this, prefer to keep the `type` handling and expose that over introducing an extra download. This can be handled from Blits by settings `type: 'compressed'` without introducing additional changes on the renderer.

PR #541, opened 2025-03-11
fix: setState dimensions undefined

pecoram:


src/core/textures/Texture.ts
```diff
@@ -259,11 +266,19 @@ export abstract class Texture extends EventEmitter {
 
     let payload: Error | Dimensions | null = null;
     if (state === 'loaded') {
-      (this.dimensions as Dimensions) = errorOrDimensions as Dimensions;
-      payload = this.dimensions;
+      if (
+        errorOrDimensions &&
```
wouterlucas, line 270: `!== null` please

src/core/textures/Texture.ts
```diff
@@ -259,11 +266,19 @@ export abstract class Texture extends EventEmitter {
...
+        !(errorOrDimensions instanceof Error) &&
```
wouterlucas, line 271: This is a critical path check and we should optimize every ounce of it.

If we can avoid an instanceof check that would be great, can we ducktype this or get this from the constructor? e.g. `errorOrDimensions.constructor === Error` would be a bit faster.

If we can't, at least not using a fallsy check will already help a little bit, eg. `(errorOrDimensions instanceof Error) === false`

src/core/textures/Texture.ts
```diff
@@ -259,11 +266,19 @@ export abstract class Texture extends EventEmitter {
...
+        (errorOrDimensions.width !== undefined ||
+          errorOrDimensions.height !== undefined)
```
wouterlucas, line 273: if its an error we don't care about width/height anymore right?

wouterlucas (changes requested):

src/core/textures/Texture.ts
pecoram: right

wouterlucas (approved):
Thanks!

PR #542, opened 2025-03-17
Feat/refactor texture coords

jfboeve:
Refactored the way textureCoords are calculated for the WebGlRenderer.

Before these were calculated every time a node was passed to the addQuad function of the renderer.

Now it's processed every time a texture is changed or when a node is created and is renderable.

### DIFF
--- a/src/core/CoreNode.ts
+++ b/src/core/CoreNode.ts
@@ -897,10 +897,6 @@ export class CoreNode extends EventEmitter {
         type: 'texture',
         dimensions,
       } satisfies NodeTextureLoadedPayload);
-
-      if (this.stage.renderer.getTextureCoords !== undefined) {
-        this.textureCoords = this.stage.renderer.getTextureCoords(this);
-      }
     }
 
     // Trigger a local update if the texture is loaded and the resizeMode is 'contain'
@@ -1261,6 +1257,14 @@ export class CoreNode extends EventEmitter {
       this.sortChildren();
     }
 
+    if (
+      this.stage.renderer.getTextureCoords !== undefined &&
+      this.textureCoords === undefined &&
+      this.isRenderable === true
+    ) {
+      this.textureCoords = this.stage.renderer.getTextureCoords(this);
+    }
+
     // If we're out of bounds, apply the render state now
     // this is done so nodes can finish their entire update loop before
     // being marked as out of bounds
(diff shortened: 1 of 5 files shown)


src/core/CoreNode.ts
```diff
@@ -1255,6 +1257,14 @@ export class CoreNode extends EventEmitter {
       this.sortChildren();
     }
 
+    if (
+      this.stage.renderer.getTextureCoords !== undefined &&
```
wouterlucas, line 1261: would there ever be a possibility for a renderer implementation to _not_ to have `getTextureCoords` implemented? 
If no, then we should just assume that's there and not check for it

src/core/CoreNode.ts
```diff
@@ -1255,6 +1257,14 @@ export class CoreNode extends EventEmitter {
...
+      this.textureCoords === undefined &&
+      this.isRenderable === true
+    ) {
+      this.textureCoords = this.stage.renderer.getTextureCoords(this);
```
wouterlucas, line 1265: why get this in the constructor? would you always require texture coords even if its an empty holder node with no dimensions?

If no, I'd suggest to do this in an update step only _if_ the node is required to get coordinates to avoid introducing extra steps on node initialization.

wouterlucas (changes requested):
Nice1 - does this make a significant impact on the benchmarks?

src/core/CoreNode.ts
jfboeve: The Canvas Renderer does not need these textureCoords.

src/core/CoreNode.ts
jfboeve: Yeah this should only be done when a node is renderable, which is why the isRenderable === true check was there, however the check was a bit extensive so I've simplified it

src/core/CoreNode.ts
```diff
@@ -1509,6 +1508,13 @@ export class CoreNode extends EventEmitter {
    */
   setRenderable(isRenderable: boolean) {
     this.isRenderable = isRenderable;
+    if (
```
wouterlucas, line 1511: mm this potentially gets called a lot, nothing bad at this point but we may need to keep an eye out for what this does to overall performance.

That's probably offsetted by the changes you've done further down the `renderQuads` pipeline 😃 - lets test this out!

wouterlucas (approved):

src/core/CoreNode.ts
jfboeve: I get what you mean. There might be other solutions but not sure if we wind up with more checks that is necessary

PR #545, opened 2025-03-26
fixed issue with shader switching

jfboeve:
Fixed an issue where switching from one shader to a different shader does not trigger the calculation of shader specific values when needed.


wouterlucas (approved):
👀

PR #548, opened 2025-03-30
Update WebGlRenderer.ts

chiefcll:
These are never undefined, they are set to false by default making this always return false


wouterlucas:
Tests seems to disagree - either there is an implicit assumption somewhere, or the type definition is off.

chiefcll:
You're welcome to look into it - 
`public parentHasRenderTexture = false;`
`rtt: false,`

Probably both are off - essentially no Ops are being reused, this change will make some be reused. Might be faster / easier to just not check and always create new render ops - but it's leading to high GC.

PR #549, opened 2025-03-30
Update CoreNode.ts

chiefcll:
Remove creating new object for every node being rendered.

### DIFF
--- a/src/core/CoreNode.ts
+++ b/src/core/CoreNode.ts
@@ -2408,7 +2408,7 @@ export class CoreNode extends EventEmitter {
     if (this.parentHasRenderTexture && !this.rtt && this.parent) {
       return this.parent.framebufferDimensions;
     }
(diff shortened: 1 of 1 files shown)


wouterlucas:
What does this solve? Is it a good idea to return the entire scope of the CoreNode instead of spawning an object with answers?

wouterlucas:
Yeah fair enough, that helps on GC pressure indeed. Looks good to me, @jfboeve ?

wouterlucas:
OK talking this through with @jfboeve and he pointed out this only needs to happen if there is a RTT parent, which would make this framebuffer dimension getter a bit off to begin with.

We'll fab up another PR to solve this higher up, we shouldn't need to pass framebuffer dimensions unless its a child of an RTT parent, which is cached. For all other nodes this isnt required at all.

Thanks for raising it!

PR #552, opened 2025-04-01
refactor: simplify isProduction check and remove unused assertions

### LINKED ISSUE #544: 🐛 [BUG] - isProductionEnv() not removed at build time
### Description

When running a production build with Vite

```
export function isProductionEnvironment(): boolean {
  return import.meta.env && import.meta.env.PROD;
}
```

Is never fully removed from the build because `import.meta.env.PROD` is true and import.meta.env must be checked at runtime.

Alternatively, if you do:

```
function isDevEnv(): boolean {
  return !!(import.meta.env && import.meta.env.DEV);
}
export const isDev = isDevEnv() || false;
```

And check for isDev (not a function) - it will be able to remove all the functions because dev is false and the const becomes false at build time. 

Currently all the assertTruthy are left in at runtime as well as isProductionEnvironement() checks. You may need to do `isDev && assertTruthy` to fully remove them

### Reproduction URL

None

### Reproduction steps

```bash
`vite build`
check output in dist directory
```

### Screenshots

```bash
![DESCRIPTION](LINK.png)
```

### Logs

```bash

```

### Browsers

Other

### Browser version

```bash
All
```

### OS

_No response_

wouterlucas:
Fixes #544 (partially) - will limit it to:

```
function assertTruthy(condition, message) {
  return;
}
```

In the production bundle. Removing it entirely is too intrusive right now as we would need to inline every `assertTruthy` check.


chiefcll:
LGTM - like the __DEV__ flag

src/utils.ts
```diff
@@ -80,7 +89,7 @@ export function assertTruthy(
   condition: unknown,
   message?: string,
 ): asserts condition {
-  if (isProductionEnvironment() === true) return;
+  if (isProductionEnvironment) return;
```
jfboeve, line 92: isProductionEnvironment === true

PR #561, opened 2025-04-16
docs: fix broken example project

minhui-foxtel:
Hi, I just found the example project is broken, hope I'm doing the right thing. Now it can be rendered like the following screenshot.

example-project


wouterlucas:
hey ! Yeah those examples are a bit overlooked, technically you would never create your custom texture like that, instead you'd use the `createTexture` texture factory API to create your own texture.

So the example is a bit too deep into the internals of the renderer and not something I would recommend to use in a real world scenario 😄

jfboeve (approved):

PR #564, opened 2025-04-18
fix detect touch test with right function name

jfboeve:
Forgot to update function name in the example test.


michielvandergeest (commented):
LGTM

Issue #567, opened 2025-05-06
🐛 [BUG] - Border of width: 0 still shows up


chiefcll:
### Description

If I create a shader with radius: 6 and border: { width: 0} - It shows a blurry border

Border width 0
<img width="494" alt="Image" src="https://github.com/user-attachments/assets/25586d8f-abee-455f-a021-ce9650397d69" />

with width 1

<img width="515" alt="Image" src="https://github.com/user-attachments/assets/6e507937-d41c-4e84-b21e-37ff7499f036" />

### Reproduction URL

none

### Reproduction steps

```bash
Create a node with radius and border width of 0
```

### Screenshots

```bash

```

### Logs

```bash

```

### Browsers

Chromium

### Browser version

```bash
Chrome desktop
```

wouterlucas:
Fixed part of #571

PR #571, opened 2025-05-09
fixed shader render issue when border is 0

jfboeve:
Fixed the webgl shaders that draw borders,  now it doing a check that when all border values are zero its skips the border calculations and draw functions.


wouterlucas (approved):

wouterlucas:
is this related to #567 ?

PR #572, opened 2025-05-12
fix only use type of webglrenderer not actual instance

jfboeve:


wouterlucas (approved):
Thanks that should do it 😃

PR #575, opened 2025-05-28
Improved update function

jfboeve:
- removed update type ScaleRotate
- reduced the amount of functions fired in update function


wouterlucas (approved):
Nice 👍

PR #588, opened 2025-06-16
fix: cleanup texture objects (v2.x)

wouterlucas:


michielvandergeest (approved):
LGTM

Tested and this seems to properly dereference texture objects when memory pressure is hit.

PR #614, opened 2025-07-22
removed contain and replaced it with maxWidth / maxHeight

jfboeve:
- Removed contain
- Introduced maxWidth / maxHeight
- also now properly loading canvas fonts
- changed update order
- overwriting some CoreNode functions to make them do less work than needed.


wouterlucas:
Whats the driver / rationale to remove contain? does `maxWidth` / `maxHeight` provide more control?
Other then its easier to understand.

Does this align with the Blits text syntax / implementation @michielvandergeest ?

wouterlucas:
nice ok - thanks for elaborating!

PR #585, opened 2025-06-10
Text engine Refactor

wouterlucas:
Text Refactor of the **Canvas Text** and **MSDF Text** renderer.

# Warning! Breaking changes

* Changes of how fonts are loaded, see instructions/examples for new loadFont API
* Removal of previously used Text Renderer / Font Classes and TS definitions
* Removal of `contain` property in favor of `maxWidth` and `maxHeight` properties

## Current status:
- [x] Stage / init cleanup (done)
- [x] Canvas text engine rewrite
- [x] SDF text engine rewrite
- [x] Text Loader rewrite
- [x] Update tests
- [x] Validation
- [x] Benchmarks

Related to https://github.com/lightning-js/renderer/issues/483

PR #615, opened 2025-07-22
Implement text baseline sdf property and changes to vertical Align

Drulokia:
### 1. textBaseline Implementation for SDF Renderer
- **Fixed textBaseline positioning**: Previously, SDF renderer incorrectly used `verticalAlign` for baseline calculations in `getStartConditions.ts`
- **Added proper textBaseline support** with all standard values:
  - `'alphabetic'` (default) - baseline at Y coordinate
  - `'top'` - top of text bounding box
  - `'middle'` - middle of text bounding box  
  - `'bottom'` - bottom of text bounding box
  - `'hanging'` - hanging baseline (96% of ascender)
  - `'ideographic'` - ideographic baseline (similar to bottom)

### 2. Changes to the verticalAlign
- **Updated `getStartConditions`** to use `textBaseline` instead of `verticalAlign` for baseline calculations
- **Enhanced `layoutText`** function to properly handle both properties independently
- **Improved vertical alignment calculations** that work correctly with textBaseline
- Default verticalAlign is now top


wouterlucas:
This too will need to be rebased/refactored into #585 as we're not going to continue with the old Text engine implementation

wouterlucas:
Superceded by new text renderer implementation - if we want to port this please raise a new PR

PR #620, opened 2025-08-11
Improved text generation check update loop text node

jfboeve:
Now checking if text is allowed to generate based on parent location / worldAlpha


wouterlucas (approved):
LGTM

PR #621, opened 2025-08-12
Synchronous font loading

wouterlucas:


jfboeve (approved):

src/core/CoreTextNode.ts
```diff
@@ -287,6 +289,19 @@ export class CoreTextNode extends CoreNode implements CoreTextNodeProps {
     }
   }
 
+  override destroy(): void {
+    this.fontHandler.stopWaitingForFont(this.textProps.fontFamily, this);
```
michielvandergeest, line 293: wondering if it's more performant to just keep the nodes in the object and skip over any nodes that have been destroyed once the font resolves, versus removing right away?

additionally, it's technically possible to change the font of a node, while the font is loading - so if we want to be super diligent we'd have to move nodes waiting for fonts to different fontFamily groups in the object, so we don't trigger the update for the previous font family while the new font family is still loading

jfboeve: I've added a flag here, that if the corenode is waiting for a font it will fire that function. I'd rather not check for every node waiting if they are destroyed or not. Also added a similar check in the set fontFamily.

PR #622, opened 2025-08-12
reinstate w/h shorthand properties

wouterlucas:
Originally L3 renderer had w/h as short hand properties - these got moved to `width` - `height`. This is reinstates the original `w` and `h` properties to match how we use them upstream.

`width` and `height` are still available but with a deprecation warning.

Events still report `width` / `height` for dimensions, do we want those to be updated as well @michielvandergeest ?

### DIFF
--- a/examples/tests/width-height-shortcuts.ts
+++ b/examples/tests/width-height-shortcuts.ts
@@ -1,0 +1,134 @@
+/*
+ * If not stated otherwise in this file or this component's LICENSE file the
+ * following copyright and licenses apply:
+ *
+ * Copyright 2025 Comcast Cable Communications Management, LLC.
+ *
+ * Licensed under the Apache License, Version 2.0 (the License);
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { ExampleSettings } from '../common/ExampleSettings.js';
+
+export async function automation(settings: ExampleSettings) {
+  await test(settings);
+  await settings.snapshot();
+}
+
+/**
+ * Tests that w/h shorthand properties work identically to width/height
+ * on CoreNodes
+ *
+ * @param settings
+ * @returns
+ */
+export default async function test(settings: ExampleSettings) {
+  const { renderer, testRoot } = settings;
+
+  // Set test area
+  testRoot.width = 600;
+  testRoot.height = 270;
+  testRoot.color = 0x222222ff;
+
+  // Create header text
+  renderer.createTextNode({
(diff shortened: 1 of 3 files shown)


chiefcll:
### **Why `width` and `height` are preferable**

* **Clarity and Self-Documentation**
  `width` and `height` are immediately obvious to anyone reading the code — including people brand-new to the project or even to programming.
  `w` and `h` require context: to you and your team it’s second nature, but to a new contributor, they could mean “weight” or “window” or something else.

* **Consistency with the JavaScript / DOM ecosystem**
  The HTML DOM, CSSOM, Canvas API, and almost every JS graphics API use `width` and `height`.
  Sticking with these terms reduces mental switching costs when moving between different parts of a codebase or different frameworks.

* **Better for tooling and searchability**
  Searching the codebase for `width` finds all relevant references — searching for `w` often returns noise (variables named `w`, temporary counters, loop vars, etc.).
  Auto-complete in IDEs is more predictable and descriptive.

* **Reduces “hidden shorthand” culture**
  Short forms encourage devs to make similar abbreviations elsewhere (`xPos` → `x`, `opacity` → `o`, etc.), leading to an API that’s harder to read for anyone who didn’t grow up inside the codebase.

Might also want to let the Plex folks know you're breaking their upstream framework as well.

wouterlucas:
Yeah good point - will let Plex know!

examples/tests/resize-mode.ts
```diff
@@ -252,16 +252,16 @@ export default async function test(settings: ExampleSettings) {
...
+          h: SQUARE_SIZE - 200,
           parent: rowNode,
           color: 0x333333ff,
           clipping: true,
         } satisfies Partial<INodeProps>;
 
         const textureNodeProps = {
-          width: containerProps.width,
-          height: containerProps.height,
+          w: containerProps.width,
```
michielvandergeest, line 263: this should probably be `containerProps.w` ?

examples/tests/resize-mode.ts
```diff
@@ -294,14 +294,14 @@ export default async function test(settings: ExampleSettings) {
           ...textureNodeProps,
           mount: 0.5,
           x: containerProps.width / 2,
```
michielvandergeest, line 296: here as well (`width` = `w`)

examples/tests/resize-mode.ts
```diff
@@ -310,16 +310,16 @@ export default async function test(settings: ExampleSettings) {
...
+          h: SQUARE_SIZE - 200,
           parent: rowNode,
           color: 0x333333ff,
           clipping: true,
         } satisfies Partial<INodeProps>;
 
         const textureNodeProps = {
-          width: containerProps.width,
-          height: containerProps.height,
+          w: containerProps.width,
```
michielvandergeest, line 321: width => w

examples/tests/resize-mode.ts
```diff
@@ -352,14 +352,14 @@ export default async function test(settings: ExampleSettings) {
           ...textureNodeProps,
           mount: 0.5,
           x: containerProps.width / 2,
```
michielvandergeest, line 354: width => w

michielvandergeest (changes requested):
Nice work ❤️ 

I found a few minor errors, possibly the ones responsible for the failing VR tests?

jfboeve (approved):

PR #639, opened 2025-09-19
feat: Allow custom timing functions for easing

WillsonHaw:
Allows specifying a timing function instead of selecting from predefined easings or a cubic-bezier ease.

I've reused the `easing` property to set the timing function, so you can specify a function or a string as before.

### DIFF
--- a/examples/tests/animation.ts
+++ b/examples/tests/animation.ts
@@ -17,12 +17,15 @@
  * limitations under the License.
  */
 
-import type { IAnimationController } from '@lightningjs/renderer';
-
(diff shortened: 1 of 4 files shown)


jfboeve:
Looks good, while you're at it could you make the timingFunction always return a number instead of number | undefined

WillsonHaw:
@jfboeve Applied your suggestion.

jfboeve (approved):

jfboeve:
Awesome @WillsonHaw, thank you for the contribution

PR #652, opened 2025-10-06
Introduce TextureError class for improved texture error handling

pecoram:


src/core/TextureMemoryManager.ts
```diff
@@ -315,9 +316,13 @@ export class TextureMemoryManager {
       });
 
       if (this.debugLogging === true || isProductionEnvironment() === false) {
-        console.warn(
-          `[TextureMemoryManager] Memory usage above critical threshold after cleanup: ${this.memUsed}`,
-        );
+        const now = Date.now();
+        if (now - this.lastCriticalWarnTime >= 500) {
```
wouterlucas, line 320: honestly - we should just hide this behind debug logging entirely. It makes no sense to spam this every frame for non-debug instances.

src/core/CoreTextureManager.ts
```diff
@@ -411,7 +415,13 @@ export class CoreTextureManager extends EventEmitter {
       this.stage.txMemManager.criticalCleanupRequested === true
     ) {
       // we're at a critical memory threshold, don't upload textures
-      texture.setState('failed', new Error('Memory threshold exceeded'));
+      texture.setState(
+        'failed',
+        new TextureError(
```
wouterlucas, line 420: can we make this an enum that resolves? instead of dealing with casting strings (and the ability to get this wrong) 

```
new TextureError(TextureError.MemoryExceeded)
```

either as a static on the TextureError class or as a separately exposed Enum.

and do a quick double lookup on the error handling side?

wouterlucas (changes requested):

src/core/TextureMemoryManager.ts
pecoram: That's already the case. If debugging is not active, it is not printed.

src/core/CoreTextureManager.ts
pecoram: Yes, this is a trick to be able to use both TextureErrorCode.CODE and “CODE.” If you make a mistake, typescript won't build. Anyway, I'll put it as an enum.

PR #654, opened 2025-10-08
fix some performance issues in v2

chiefcll:
1. Back ported the isProductionEnvironment so AssertTruthy can be tree shaken
2. Back port exporting of utils
3. Optimize WebGL loop as Shaders are ALWAYS defined on CoreNode and shaderProps as well.

Requires __DEV__ to be added to vite config.


chiefcll:
Update the VRT and fixed up merge conflict

PR #656, opened 2025-10-17
fixed gradient effects

jfboeve:
Someone found out that bright colors caused some issues displaying the wrong color.

In order to fix this, I changed the fragment shader parts for both effects, they should now properly render the right color.


michielvandergeest (approved):
LGTM

PR #672, opened 2025-12-05
Fix text width handling in inspector

pecoram:
- Adjust the width and height properties to return 'auto' instead of '1px' for zero values.

### DIFF
--- a/src/main-api/Inspector.ts
+++ b/src/main-api/Inspector.ts
@@ -575,7 +575,10 @@ export class Inspector {
     // special case for text
     if (property === 'text') {
       div.innerHTML = String(value);
-      div.style.visibility = 'hidden';
+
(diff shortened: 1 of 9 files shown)


src/main-api/Inspector.ts
```diff
@@ -52,16 +52,14 @@ const stylePropertyMap: {
...
+      return { prop: 'width', value: 'auto' };
     }
 
     return { prop: 'width', value: `${w}px` };
   },
   height: (h) => {
     if (h === 0) {
-      // Set to 1px instead of 0px so visibility checks (e.g. Playwright) keep treating the node as visible
-      return { prop: 'height', value: '1px' };
+      return { prop: 'height', value: 'auto' };
```
wouterlucas: will auto still ensure its visible? even if it has no width set?

PR #677, opened 2025-12-19
fix: add safety against no ctxTexture

pawelocode:
This change fixes an issue seen on specific devices since V2.20.0.

Im not sure if this is the best approach but hopefully this PR starts the conversation

### DIFF
--- a/src/core/renderers/webgl/WebGlCoreRenderer.ts
+++ b/src/core/renderers/webgl/WebGlCoreRenderer.ts
@@ -294,6 +294,12 @@ export class WebGlCoreRenderer extends CoreRenderer {
       ctxTexture = texture.ctxTexture as WebGlCoreCtxTexture;
     } else {
       ctxTexture = texture.ctxTexture as WebGlCoreCtxTexture;
+      if (ctxTexture === undefined) {
+        ctxTexture = this.stage.defaultTexture?.ctxTexture as WebGlCoreCtxTexture;
+        console.warn(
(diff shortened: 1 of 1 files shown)


michielvandergeest:
this PR originated from an issue that we could only reproduce on the slowest Sky Q device in combination with compressed textures (which may take longer to download because of the file size). So some slowness on devices could be responsible for this error

@pawelocode can you comment on the following question: _Does the rendering have a 1 or 2 frame artifact and then auto corrects or is this always going to be broken (ie the texture never got created versus its slow)?_

when you tested did the image ultimately correct itself? or remained broken forever? Even with this fix

yananazarenko417:
just to add to my previous comment: the app fails even on the high end device (v 2.20), so I am not sure it has anything to do with the slowness of the texture creation. It does not happen on all devices though, I had the app failing completely (black screen) on 2 Samsungs but not on LG. It does clearly happen as soon as I have an asset that has failed to load in the viewport. The app fails cause it still tries to read the `txCoordX1`, although the texture is undefined. So clearly some safety missing
Works just fine and never fails on all the versions before

didn't test the fix from this MR yet

wouterlucas:
OK - would be good to find the root cause instead of adding a safety that only prevents the pipeline from crashing.
Agree that's likely a good long term stability thing to have, but this might mask uncovering what the root cause is.

@pawelocode I'll reach out on slack to see if we can debug this together with @jfboeve

PR #681, opened 2025-12-28
Autosize node based on children

wouterlucas:
Add autosize to a node to dynamically update its size based on children's matrix3d.

* Reuse of `autosize` property
* If node has a texture, that behavior will be prioritized (to not break existing behavior)
* If node does not have a texture, w/h are automatically adjusted to envelop children

notes:
* Does not grow "backward" so only beyond its own x,y anchor left + down
* Although optimized, this is not "for free" as it will involve additional calculations

PR includes unit tests and VRT testcase.

Useful for Touch/Point usecases where you want the parent to be clickable and may provide utility for flexbox usage. 

image

### DIFF
--- a/src/core/CoreNode.ts
+++ b/src/core/CoreNode.ts
@@ -57,7 +57,7 @@ import type { IAnimationController } from '../common/IAnimationController.js';
 import { CoreAnimation } from './animations/CoreAnimation.js';
 import { CoreAnimationController } from './animations/CoreAnimationController.js';
 import type { CoreShaderNode } from './renderers/CoreShaderNode.js';
-import { Autosizer } from './Autosizer.js';
+import { AutosizeMode, Autosizer } from './Autosizer.js';
 import {
   bucketSortByZIndex,
   incrementalRepositionByZIndex,
@@ -2638,7 +2638,7 @@ export class CoreNode extends EventEmitter {
     if (oldTexture) {
       this.unloadTexture();
       if (this.autosizer !== null && value === null) {
-        this.autosizer.setMode(0); // Set to children size mode
+        this.autosizer.setMode(AutosizeMode.Children); // Set to children size mode
       }
     }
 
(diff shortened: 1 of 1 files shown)


src/core/CoreNode.ts
```diff
@@ -1131,12 +1177,14 @@ export class CoreNode extends EventEmitter {
...
-        UpdateType.RecalcUniforms;
+      updateType |= UpdateType.RenderState | UpdateType.RecalcUniforms;
       updateParent = hasParent;
-      childUpdateType |= UpdateType.Global;
+
+      //only propagate children updates if not autosizing
+      if ((updateType & UpdateType.Autosize) === 0) {
+        updateType |= UpdateType.Children;
+        childUpdateType |= UpdateType.Global;
+      }
```
wouterlucas: why is this check there to prevent updates to children?

src/core/CoreNode.ts
```diff
@@ -1181,6 +1229,15 @@ export class CoreNode extends EventEmitter {
       this.updateIsRenderable();
     }
 
+    // Handle autosize updates when children transforms change
+    if (
+      updateType & UpdateType.Global &&
+      this.isRenderable === true &&
+      this.parentAutosizer !== null
+    ) {
+      this.parentAutosizer.patch(this);
```
wouterlucas: could do `patch(this._id)`

wouterlucas (commented):
nice - looks good, minor remarks.

it would make sense to test this with some large deeply nested child trees. Just to profile what the impact is of having a lot of children/deeply nested children when navigating around.

src/core/Autosizer.ts
```diff
@@ -0,0 +1,219 @@
...
+  All = 2,
+}
+
+const applyDimensions = (
+  node: CoreNode,
+  carryOver: boolean,
+  w: number,
+  h: number,
+) => {
+  if (carryOver === true) {
```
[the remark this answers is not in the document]
jfboeve: forget about this, I removed it and changed a bit of logic to avoid unnessecary calcs.

PR #685, opened 2025-12-29
CanvasTextRenderer to use settings dpr

estobbart:
The actual dpr is established higher up in the renderer, resolving the incoming settings & fallback to window.

should the `|| 1` still happen here? or should that also resolve earlier in the render creation?


wouterlucas:
no need to keep the `|| 1` in there - this is resolved higher up in `Renderer.ts`

jfboeve (approved):

Issue #670, opened 2025-12-03
🐛 [BUG] - MountX not centering correctly [Beta16]


chiefcll:
### Description

Given a node with some text, after updating the width of the container with mountX = 0.5 it is not correctly centering.

<img width="410" height="160" alt="Image" src="https://github.com/user-attachments/assets/1ae03bca-e0a6-4ea0-9ab2-94460191f9f8" />

### Reproduction URL

none

### Reproduction steps

If I use the inspector after its rendered and change the width or the mountX again it will update to the correct position. Not sure where the timing issue is with it not doing it correctly the first time.

<img width="396" height="105" alt="Image" src="https://github.com/user-attachments/assets/3626fb89-443f-4779-aa96-19d3b7572faf" />

Fixed via inspector setting same values

### Screenshots

```bash
![DESCRIPTION](LINK.png)
```

### Logs

```bash

```

### Browsers

Chromium

### Browser version

```bash
Beta-16
```

PR #687, opened 2026-01-03
fix updating nodes during update loop

chiefcll:
Resetting update type should be done at top of the loop as nodes can change values during the render cycle and can be redrawn on the next frame.

### DIFF
--- a/src/core/CoreNode.ts
+++ b/src/core/CoreNode.ts
@@ -1047,6 +1047,10 @@ export class CoreNode extends EventEmitter {
     let childUpdateType = this.childUpdateType;
     let updateParent = false;
 
+    // reset update type
+    this.updateType = 0;
+    this.childUpdateType = 0;
+
     if (updateType & UpdateType.Local) {
       this.updateLocalTransform();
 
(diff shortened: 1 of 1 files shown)


chiefcll:
Fixes #670

jfboeve:
There's an issue with this fix as it does not go through the vrt till the end.

src/core/CoreNode.ts
```diff
@@ -1204,11 +1208,6 @@ export class CoreNode extends EventEmitter {
       parent.setUpdateType(UpdateType.ZIndexSortedChildren);
     }
 
-    if (this.renderState === CoreNodeRenderState.OutOfBounds) {
-      updateType &= ~UpdateType.RenderBounds; // remove render bounds update
-      return;
-    }
-
```
wouterlucas: Why are we removing this? This will force part of the tree that are out of bounds to continue to be processed, which essentially is wasteful as those parts aren't visible (hence take performance).

src/core/CoreNode.ts
```diff
@@ -1047,6 +1047,10 @@ export class CoreNode extends EventEmitter {
     let childUpdateType = this.childUpdateType;
     let updateParent = false;
 
+    // reset update type
+    this.updateType = 0;
+    this.childUpdateType = 0;
```
wouterlucas: How does this help exactly?

wouterlucas (changes requested):
Is your initial creation out of bounds and is it not being processed correctly before you put it on screen?

src/core/CoreNode.ts
chiefcll: This snippit makes lines 1271 unreachable. TS warns about it you're just using this.renderState rather than renderState. This prevents nodes from updating that were out of bounds and coming into bounds.

PR #690, opened 2026-01-05
inspector boolean

estobbart:
the null coalescing happens earlier when the settings goes through its normalizing step


wouterlucas:
I think the problem is that inspector can be provided as a class? to be able to keep it tree shakeable...
This seems to break that.

PR #692, opened 2026-01-06
fix rounding to be smoother on retina devices and smooth correctly

chiefcll:
Fixes rounding from:

Screenshot 2026-01-04 at 9 41 42 PM

To:

image


wouterlucas (approved):
nice - thanks!

PR #694, opened 2026-01-07
(v3) Texture Management / GC

jfboeve:
Ported Texture Refactor to v3.

Includes changes from PR's:
#647 
#649 
#650 
#651 
#652 
#653 

Also improved the flow regarding resolving props for texture creation. In the old version it happens twice for each unique texture, now it does it only once.


wouterlucas (approved):
LGTM - thanks for porting this from v2!

PR #695, opened 2026-01-07
optimize the update loop

chiefcll:
See comments in line - #687 can be closed out for this PR


src/core/CoreNode.ts
```diff
@@ -1038,10 +1046,6 @@ export class CoreNode extends EventEmitter {
    * @param delta
    */
   update(delta: number, parentClippingRect: RectWithValid): void {
-    if (this.updateType === UpdateType.None) {
-      return;
-    }
```
chiefcll: This is not needed as the child for loop checks this.

chiefcll:
Also I believe the test is broken as the code was broken before...

Expected says red rectangle is out of bounds, when it is In bounds.
viewport-events-17-expected

With the update, the red rectangle is in the viewport.
viewport-events-17-actual

PR #696, opened 2026-01-07
Add border gap and color

chiefcll:
Update Rounded & Border shaders to add border-gap and border-gapColor properties to support offset borders and fill colors inside the gap.

This PR allows gap: 6,  gap: 0, and gap: -6 to allow inset

Screenshot 2026-01-07 at 1 25 21 PM
Screenshot 2026-01-07 at 1 25 36 PM
Screenshot 2026-01-07 at 1 39 53 PM


ftardif:
We are very interested in seeing this feature supported natively in the renderer. The 'focus ring' on a selected component is a foundational pattern in modern TV UI design. Having this included and optimized at the renderer level would be a significant win for both performance and developer experience.

src/core/shaders/templates/RoundedWithBorderAndShadowTemplate.ts
```diff
@@ -23,13 +23,30 @@ import { getShadowProps, type ShadowProps } from './ShadowTemplate.js';
...
 export type RoundedWithBorderAndShadowProps = RoundedProps &
   PrefixedType<BorderProps, 'border'> &
-  PrefixedType<ShadowProps, 'shadow'>;
+  PrefixedType<ShadowProps, 'shadow'> & {
+    /**
+     * Gap between the border and the content
+     *
+     * @default 0
+     */
+    'border-gap': number;
```
girouxc: should we add it directly to BorderProps instead ?

PR #710, opened 2026-01-14
Fixes Canvas Render

jfboeve:
- Fixed an issue where normal nodes without specific textures or shaders were not being drawn. Found out a problem with hasColorProps not being set on create, this is needed for the renderable check.
- Refactored some code in CanvasRenderer where checks are not needed (or simply are done again, after they've been done earlier in the render pipeline).

### DIFF
--- a/src/core/CoreNode.ts
+++ b/src/core/CoreNode.ts
@@ -764,6 +764,9 @@ export class CoreNode extends EventEmitter {
   constructor(readonly stage: Stage, props: CoreNodeProps) {
     super();
     const p = (this.props = {} as CoreNodeProps);
+    //inital update type
+    let initialUpdateType =
+      UpdateType.Local | UpdateType.RenderBounds | UpdateType.RenderState;
 
     // Fast-path assign only known keys
     p.x = props.x;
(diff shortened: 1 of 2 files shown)


src/core/CoreNode.ts
```diff
@@ -773,8 +776,8 @@ export class CoreNode extends EventEmitter {
     p.alpha = props.alpha;
     p.autosize = props.autosize;
     p.clipping = props.clipping;
-    p.color = props.color;
 
+    p.color = props.color;
```
wouterlucas: 🫣

jfboeve: just ordering color props together xD

src/core/CoreNode.ts
```diff
@@ -784,6 +787,22 @@ export class CoreNode extends EventEmitter {
...
+      props.colorLeft > 0 ||
+      props.colorRight > 0 ||
+      props.colorTl > 0 ||
+      props.colorTr > 0 ||
+      props.colorBl > 0 ||
+      props.colorBr > 0
+    ) {
+      this.hasColorProps = true;
+      initialUpdateType |= UpdateType.PremultipliedColors;
+    }
```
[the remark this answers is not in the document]
jfboeve: could not find the setting of hasColorProps outside the setters, so i guess it must have disappeared somewhere, didn't run into this issue until checking canvas

PR #712, opened 2026-01-15
Direct texture creation for ION

wouterlucas:
@estobbart would need to figure out how to get the true Texture/Image dimensions from ION - should we enhance that on the `createImageBitmap` response or do something more specific?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@lightningjs/renderer",
-  "version": "3.0.0-beta19",
+  "version": "3.0.0-beta20",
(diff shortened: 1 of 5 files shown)


wouterlucas:
> I don't necessarily have specifics, only to mimic something similar to LNG2.platform.loadSrcTexture & LNG2.platform.uploadGlTexture

OK - moving this to another PR, going to redo the whole Platform stuff where you can completely abstract the createImage, fetchImage and GL wrapper to make it work for ION. Will tag you there.

PR #724, opened 2026-01-21
disable gl.getError check for production

chiefcll:
performing this check has a lot of overhead for each image.


chiefcll:
I think it's important in Dev to know if it worked. But in prod - won't matter. I'll look into moving it up...

chiefcll:
done

jfboeve (approved):

PR #725, opened 2026-01-21
cache key value for shader nodes

chiefcll:
This is currently being generated on every frame - caching the result saves a lot of processing


src/core/renderers/CoreShaderNode.ts
```diff
@@ -96,6 +96,10 @@ export class CoreShaderNode<Props extends object = Record<string, unknown>> {
   protected node: CoreNode | null = null;
   readonly time: CoreShaderType['time'] = undefined;
   update: (() => void) | undefined = undefined;
+  protected _valueKeyCache = '';
+  protected _valueKeyDirty = true;
+  protected _lastW = 0;
+  protected _lastH = 0;
```
jfboeve: can you make these new additions private

PR #730, opened 2026-01-30
added reset parameter to stop function Animation controller

jfboeve:
Added a reset parameter to the CoreAnimationController stop function, by default it's true so there are no regressions. 

The reason for this change is that currently if you use the stop function the values will reset to their start values which is not handy if you want to chain the same value to a different animation.


michielvandergeest (approved):
LGTM!

PR #740, opened 2026-02-12
feat CoreFont & CoreFontManager

jfboeve:
- Added CoreFont
- Reintroduced Fontmanager
- Unique font family names
- You can change fontfamily during runtime
- removed textRendererOverride (family names are unique now)


wouterlucas:
I feel we stripped this out to reduce layers, now reintroducing this again :face_with_peeking_eye: 

why do we need a font manager exactly?
is changing font families really a need during runtime?
does this slow stuff down again or is it neglible?

PR #750, opened 2026-03-05
Implement exponential backoff for texture loading retries

pecoram:
Introduce an exponential backoff strategy for texture loading retries to improve resource management and reduce immediate retry attempts after failures.

### DIFF
--- a/src/core/Stage.ts
+++ b/src/core/Stage.ts
@@ -344,6 +344,9 @@ export class Stage {
     this.txManager.frameTime = newFrameTime;
     this.txMemManager.frameTime = newFrameTime;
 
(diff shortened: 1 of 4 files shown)


src/core/Stage.ts
```diff
@@ -344,6 +344,9 @@ export class Stage {
     this.txManager.frameTime = newFrameTime;
     this.txMemManager.frameTime = newFrameTime;
 
+    // Process texture retries with exponential backoff
+    this.txManager.processTextureRetries();
```
wouterlucas: wouldnt it be better to make this part of `processSome` to put a time budget on it?

PR #764, opened 2026-03-19
fix: prevent initial text position jump

pecoram:
When a Text node is inside an autosized parent with mount enabled, the text appears in the wrong position for 1 frame and then snaps into place.

This is an intentional tradeoff in favor of first-frame visual stability: it shifts some work earlier during node creation to avoid visible repositioning.

### DIFF
--- a/src/core/CoreTextNode.ts
+++ b/src/core/CoreTextNode.ts
@@ -78,38 +78,25 @@ export class CoreTextNode extends CoreNode implements CoreTextNodeProps {
   private _type: 'sdf' | 'canvas' = 'sdf'; // Default to SDF renderer
 
   /**
-   * Only texts whose transform depends on measured dimensions need eager layout
-   * to avoid an initial position jump.
+   * Only texts whose parent transform/layout depends on measured dimensions
+   * need eager layout to avoid an initial position jump.
    */
   private requiresLayoutForInitialTransform(props: CoreTextNodeProps): boolean {
+    const parent = props.parent;
+    if (parent === null) {
+      return false;
+    }
+
     if (
-      props.mountX !== 0 ||
-      props.mountY !== 0 ||
-      props.pivotX !== 0 ||
-      props.pivotY !== 0
+      parent.mountX !== 0 ||
+      parent.mountY !== 0 ||
+      parent.pivotX !== 0.5 ||
+      parent.pivotY !== 0.5
     ) {
       return true;
     }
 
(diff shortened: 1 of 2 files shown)


wouterlucas (approved):
LGTM

src/core/CoreTextNode.ts
```diff
@@ -77,6 +77,28 @@ export class CoreTextNode extends CoreNode implements CoreTextNodeProps {
 
   private _type: 'sdf' | 'canvas' = 'sdf'; // Default to SDF renderer
 
+  /**
+   * Only texts whose parent transform/layout depends on measured dimensions
+   * need eager layout to avoid an initial position jump.
+   */
+  private requiresLayoutForInitialTransform(props: CoreTextNodeProps): boolean {
+    const parent = props.parent;
+    if (parent === null) {
```
wouterlucas, line 86: You are already doing a parent is null check on the prior if check:
`(props.forceLoad === true || props.parent !== null) &&`

src/core/CoreTextNode.ts
```diff
@@ -77,6 +77,28 @@ export class CoreTextNode extends CoreNode implements CoreTextNodeProps {
...
+      return false;
+    }
+
+    if (
+      parent.mountX !== 0 ||
+      parent.mountY !== 0 ||
+      parent.pivotX !== 0.5 ||
+      parent.pivotY !== 0.5
+    ) {
+      return true;
```
wouterlucas, line 96: why are we returning true when the mount and pivot is being set?

src/core/CoreTextNode.ts
```diff
@@ -92,6 +114,16 @@ export class CoreTextNode extends CoreNode implements CoreTextNodeProps {
...
 
+    if (
+      (props.forceLoad === true || props.parent !== null) &&
+      this.requiresLayoutForInitialTransform(props) === true &&
+      this.fontHandler.isFontLoaded(this.textProps.fontFamily) === true
+    ) {
+      const resp = this.textRenderer.renderText(this.textProps);
+      this.handleRenderResult(resp);
+      this._layoutGenerated = true;
+    }
```
wouterlucas, line 125: the only downside to this is it will also always render text+layouting even if the text is not in visible bounds or is alpha'd out

you may potentially be generating a lot of text that isnt even required to be rendered, maybe calling `allowTextGeneration()` would help avoid that?

wouterlucas (changes requested):

PR #775, opened 2026-04-23
fixed hasValidText check to always return a boolean

jfboeve:
when using and empty string the check did not return a boolean


michielvandergeest (dismissed):

michielvandergeest (dismissed):

michielvandergeest (approved):
still LGTM!

PR #777, opened 2026-04-28
feat: add dynamic sentence text test example

wouterlucas:


milesnash (approved):
Works on STB with the following:

QtWebEngine/5.9.7 Chrome/56.0.2924.122

PR #779, opened 2026-05-01
Styleless canvas support

estobbart:
Some canvas' just have no style


wouterlucas (approved):
That's not very classy.

PR #781, opened 2026-05-02
feat: implement renderer close()

wouterlucas:


src/core/platforms/web/WebPlatform.ts
```diff
@@ -104,6 +105,7 @@ export class WebPlatform extends Platform {
     let lastFrameTime = 0;
 
     const runLoop = (currentTime: number = 0) => {
+      if (this.stopped) return;
       const targetFrameTime = stage.targetFrameTime;
```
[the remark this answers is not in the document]
wouterlucas: @copilot apply changes based on [this feedback](https://github.com/lightning-js/renderer/pull/781#discussion_r3177127108)

wouterlucas:
@copilot Stage.destroy() aims to fully tear down resources, but it currently leaves some internal references/timers alive:

    TextureMemoryManager can start a setInterval when debugLogging is enabled, and destroy() does not clear it.
    CoreTextureManager retains strong references in its internal upload queue (queued textures), and those aren’t cleared here.
    Consider adding explicit destroy()/reset() methods on these managers (clearing intervals/queues/caches) and calling them from Stage.destroy() so a closed renderer can be GC’d even if the RendererMain object is still referenced.

can we apply the suggested above changes?

package.json
```diff
@@ -60,6 +60,7 @@
     "concurrently": "^8.2.2",
     "eslint": "^9.36.0",
     "eslint-config-prettier": "^8.10.2",
+    "happy-dom": "17.6.3",
```
jfboeve, line 63: Where is this being used?

wouterlucas: unit tests - happy dom gives a vitest dom objects so the tests can mimick running in a browser

jfboeve: ok

jfboeve (dismissed):

PR #783, opened 2026-05-03
feat: implement Sequential Animation API

wouterlucas:
adds a new `animationSequence()` method to the renderer's node API that brings Lightning 2's keyframe-based animation syntax to Lightning 3, without replacing or modifying the existing `animate()` API.

This was requested by @DouweCnossen who relies on chaining multiple animations together with stepping, repeat, and keyframe mid-stops — functionality that was not previously available in L3 and is currently doing that through the reactivity system of Blits. Whilst possible, doing this inside the renderer is far more efficient.

@michielvandergeest lets discuss this, either we expose this through an additional syntax on Blits. I've kept the original animation API there, but we could also consider replacing it or making the animation controller an export that you provide on setup to bind (thus making it tree shakeable). 

## What changed

A new `animationSequence()` method is available on all nodes. It accepts one or more animation steps, each describing which properties to animate and how, using a keyframe map where the keys are progress fractions between 0 and 1. Steps are played end-to-end in order. The method returns a standard `IAnimationController`, so it integrates with all existing animation control patterns including `start`, `stop`, `pause`, `restore`, and `waitUntilStopped`.

Notable capabilities of the new API:

- Multiple properties can be animated simultaneously within a single step
- Each step can have its own duration, delay, easing, and repeat count
- Keyframe maps support any number of intermediate stops between 0 and 1
- A step can repeat a fixed number of times or loop indefinitely before the sequence advances
- Keyframe values support an optional smoothness scalar that blends between linear and ease-in-out interpolation on a per-stop basis
- Calling `restore()` resets all animated properties to the values they had when the sequence was constructed

# Examples

**Basic two-step sequence — slide then fade:**

```ts
node.animationSequence([
  {
    duration: 500,
    actions: [{ p: 'x', v: { 0: 0, 1: 800 } }],
  },
  {
    duration: 300,
    actions: [{ p: 'alpha', v: { 0: 1, 1: 0 } }],
  },
]).start();
```

**Multiple properties animated simultaneously within one step:**

```ts
node.animationSequence({
  duration: 800,
  easing: 'ease-in-out',
  actions: [
    { p: 'x',     v: { 0: 0,   1: 600 } },
    { p: 'alpha', v: { 0: 0,   1: 1   } },
    { p: 'scale', v: { 0: 0.8, 1: 1   } },
  ],
}).start();
```

**Three-keyframe bounce with a mid-stop:**

```ts
node.animationSequence({
  duration: 900,
  actions: [{ p: 'x', v: { 0: 0, 0.5: 800, 1: 0 } }],
}).start();
```

**Step that repeats before the sequence continues:**

```ts
node.animationSequence([
  {
    duration: 300,
    repeat: 2,   // plays 3 times total
    actions: [{ p: 'alpha', v: { 0: 1, 0.5: 0.2, 1: 1 } }],
  },
  {
    duration: 500,
    actions: [{ p: 'x', v: { 0: 0, 1: 400 } }],
  },
]).start();
```

**Value smoothing with `{ v, s }` — s=0 gives ease-in-out per segment:**

```ts
node.animationSequence({
  duration: 600,
  actions: [{
    p: 'x',
    v: {
      0:   { v: 0,   s: 0 },
      0.5: { v: 400, s: 0 },
      1:   { v: 800, s: 0 },
    },
  }],
}).start();
```

**Using `waitUntilStopped` to chain follow-up work:**

```ts
const ctrl = node.animationSequence([
  { duration: 400, actions: [{ p: 'x', v: { 0: 0, 1: 400 } }] },
  { duration: 400, actions: [{ p: 'alpha', v: { 0: 1, 1: 0 } }] },
]);

ctrl.start();
await ctrl.waitUntilStopped();
// sequence fully complete — safe to clean up or trigger next action
node.parent = null;
```


wouterlucas:
dropping this in favor for external animation engines

PR #784, opened 2026-05-04
Minor RTT pref fixes + benchmark

wouterlucas:


src/core/renderers/webgl/WebGlRenderer.ts
```diff
@@ -474,54 +474,53 @@ export class WebGlRenderer extends CoreRenderer {
...
     let currentNode: CoreNode = node;
-    while (currentNode) {
-      if (!currentNode.parent) {
-        break;
-      }
-
-      const parentIndex = this.rttNodes.indexOf(currentNode.parent);
-      if (parentIndex !== -1) {
-        // Found an RTT parent in the list; set insertIndex to place node before the parent
+    while (currentNode.parent !== null && currentNode.parent !== undefined) {
```
jfboeve: parent should always be null or core node, so maybe the undefined check isn't required

src/core/renderers/webgl/WebGlRenderer.ts
```diff
@@ -474,54 +474,53 @@ export class WebGlRenderer extends CoreRenderer {
   private insertRTTNodeInOrder(node: CoreNode) {
     let insertIndex = this.rttNodes.length; // Default to the end of the array
 
+    // Build a one-shot index map so all lookups below are O(1) instead of O(n).
+    const rttNodes = this.rttNodes;
+    const indexMap = new Map<CoreNode, number>();
+    for (let i = 0; i < rttNodes.length; i++) {
+      indexMap.set(rttNodes[i]!, i);
```
jfboeve: recommend using rttNodes[i].id as key

jfboeve (approved):

PR #794, opened 2026-05-15
fix: make CanvasFontHandler compatible with FontHandler

pecoram:


src/core/text-rendering/CanvasFontHandler.ts
```diff
@@ -136,7 +136,7 @@ export const getFontFamilies = (): FontFamilyMap => {
  */
 export const init = (
   c: CanvasRenderingContext2D | OffscreenCanvasRenderingContext2D,
-  mc: CanvasRenderingContext2D | OffscreenCanvasRenderingContext2D,
+  mc?: CanvasRenderingContext2D | OffscreenCanvasRenderingContext2D,
```
guilhermesimoes, line 139:
```suggestion
  mc = c,
```
This works

src/core/text-rendering/CanvasFontHandler.ts
```diff
@@ -149,7 +149,7 @@ export const init = (
   }
 
   context = c;
-  measureContext = mc;
+  measureContext = mc ?? c;
```
guilhermesimoes: Given we're not dealing with numbers or booleans, there's no reason to use `??` when we can use `||`. `??` transpiles to more complicated code, so `||` is preferable: https://www.typescriptlang.org/play/?#code/MYewdgzgLgBAhjAvDKAnArgUwNwChSSwBGS8MA-OTAGZwA2EO+40MwpCAPpzfY9kA

In any case, if we use a default param like I suggested in my other comment, then this isn't even needed:
```suggestion
  measureContext = mc;
```

jfboeve (approved):

PR #795, opened 2026-05-18
Improve inspector

guilhermesimoes:
Each commit improves its own thing, so I recommend reviewing commit by commit.

- Fix inspector `parent` handling. This includes:
  - Not appending to the DOM `div`s of childs created without a parent, and assigned a parent _afterwards_.
  - Not removing from the DOM `div`s of childs after they're removed from their parent.
- Search up the entire proto tree for getters and setters to override (needed when extending `CoreNode`).
- Show component class name in the inspector (very useful when extending `CoreNode`).

### DIFF
--- a/src/main-api/Inspector.ts
+++ b/src/main-api/Inspector.ts
@@ -1101,7 +1101,7 @@ export class Inspector {
     value: any,
     props: CoreNodeProps | CoreTextNodeProps,
   ) {
-    if (this.root === null || value === undefined || value === null) {
+    if (this.root === null || value === undefined) {
       return;
     }
 
@@ -1109,17 +1109,18 @@ export class Inspector {
      * Special case for parent property
      */
     if (property === 'parent') {
-      const parentId: number = value.id;
-
-      // only way to detect if the parent is the root node
-      // if you are reading this and have a better way, please let me know
-      if (parentId === 1) {
-        this.root.appendChild(div);
-        return;
+      if (value) {
+        const parentId: number = value.id;
+        // only way to detect if the parent is the root node
+        // if you are reading this and have a better way, please let me know
+        if (parentId === 1) {
+            this.root.appendChild(div);
+            return;
+        }
+        value.div.appendChild(div);
+      } else {
+        div.parentNode?.removeChild(div);
       }
-
-      const parent = document.getElementById(parentId.toString());
-      parent?.appendChild(div);
       return;
     }
 
--- a/src/main-api/Inspector.ts
+++ b/src/main-api/Inspector.ts
@@ -1111,14 +1111,12 @@ export class Inspector {
      */
     if (property === 'parent') {
       if (value) {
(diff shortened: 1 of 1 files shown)


src/main-api/Inspector.ts
```diff
@@ -1101,25 +1102,26 @@ export class Inspector {
...
-        this.root.appendChild(div);
-        return;
+      if (value) {
+        const parentId: number = value.id;
+        // only way to detect if the parent is the root node
+        // if you are reading this and have a better way, please let me know
+        if (parentId === 1) {
+            this.root.appendChild(div);
+            return;
+        }
```
guilhermesimoes: Btw I have another way of doing this:
```suggestion
        // detect if the parent is the root node
        if (value === value.stage.root) {
            this.root.appendChild(div);
            return;
        }
```
What do you think? Would you prefer this?

src/main-api/Inspector.ts
```diff
@@ -1101,25 +1102,26 @@ export class Inspector {
...
      */
     if (property === 'parent') {
-      const parentId: number = value.id;
-
-      // only way to detect if the parent is the root node
-      // if you are reading this and have a better way, please let me know
-      if (parentId === 1) {
-        this.root.appendChild(div);
-        return;
+      if (value) {
```
jfboeve, line 1113: value !== null

src/main-api/Inspector.ts
jfboeve: I do like this way better, but to prune it a bit more I would suggest to do:

value.id === value.stage.root.id

jfboeve (changes requested):

src/main-api/Inspector.ts
guilhermesimoes: Done!

src/main-api/Inspector.ts
guilhermesimoes: Does it matter if the value is `null` or some other falsy value?

This implementation matches the others:

https://github.com/lightning-js/renderer/blob/8ca0babdde64b4a894d827319caf83d67df9cf8d/src/main-api/Inspector.ts#L1141

https://github.com/lightning-js/renderer/blob/8ca0babdde64b4a894d827319caf83d67df9cf8d/src/main-api/Inspector.ts#L1154

The top guard also protects against `undefined`:

https://github.com/lightning-js/renderer/blob/63671a717f80165065ce13bff38409d4e5c1f8d5/src/main-api/Inspector.ts#L1098-L1107

jfboeve: uhh yeah you're right

jfboeve (approved):

jfboeve:
Looks good! Thanks for contributing

src/main-api/Inspector.ts
guilhermesimoes: Thank you!

PR #796, opened 2026-05-18
Fix shader setter not updating `hasShaderUpdater` and `hasShaderTimeFn`

guilhermesimoes:
Fix crash:

> TypeError: this.shader.update is not a function

 caused by doing:

```ts
this.bar = renderer.createNode({
    w: BAR_WIDTH,
    h: BAR_HEIGHT,
    color: 0x00000000,
    shader: renderer.createShader('border', { w: 4, color: 0x000000cc }),
});

setTimeout(() => {
  this.bar.shader = renderer.createShader('default');
}, 5000);
```

Now, the properties `hasShaderUpdater` and `hasShaderTimeFn` are always updated when the shader is set to `null` or to the `default` shader.

### DIFF
--- a/src/core/CoreNode.test.ts
+++ b/src/core/CoreNode.test.ts
@@ -26,7 +26,7 @@ import { type TextureOptions } from './CoreTextureManager.js';
 import { createBound } from './lib/utils.js';
 import { ImageTexture } from './textures/ImageTexture.js';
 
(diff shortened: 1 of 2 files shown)


jfboeve (approved):

jfboeve:
Nice find!

PR #797, opened 2026-06-01
Fixed animation stutter

jfboeve:
Fixed / reduced animation stutter on low-end devices.

Problem was cause by array creation every frame for each animation, which in return triggered the Garbage Collector causing stutters.

### DIFF
--- a/src/core/animations/CoreAnimation.ts
+++ b/src/core/animations/CoreAnimation.ts
@@ -50,7 +50,7 @@ type PropValuesMap = {
 let animationIdCounter = 0;
 
 export class CoreAnimation extends EventEmitter {
-  public readonly id: number = ++animationIdCounter | 0;
+  public readonly id: number = ++animationIdCounter;
   public settings: AnimationSettings;
   private progress = 0;
   private delayFor = 0;
--- a/src/core/animations/CoreAnimation.ts
+++ b/src/core/animations/CoreAnimation.ts
@@ -76,7 +76,7 @@ export class CoreAnimation extends EventEmitter {
           start:
             (node[
               key as keyof Omit<CoreNodeAnimateProps, 'shaderProps'>
-            ] as number) || 0,
+            ] as number) ?? 0,
           target: props[
             key as keyof Omit<CoreNodeAnimateProps, 'shaderProps'>
           ] as number,
--- a/src/core/animations/CoreAnimation.ts
+++ b/src/core/animations/CoreAnimation.ts
@@ -296,7 +296,7 @@ export class CoreAnimation extends EventEmitter {
     }
 
     if (this.progress < 1) {
-      this.emit('tick', { progress: this.progress });
+      this.emit('tick');
     }
 
     if (this.progress === 1) {
--- a/src/core/animations/CoreAnimation.ts
+++ b/src/core/animations/CoreAnimation.ts
@@ -68,32 +68,30 @@ export class CoreAnimation extends EventEmitter {
 
     for (const key in props) {
       if (key !== 'shaderProps') {
-        if (this.propValuesMap.props === null) {
-          this.propValuesMap.props = { keys: [], values: [] };
+        if (this.propValuesMap['props'] === null) {
+          this.propValuesMap['props'] = { keys: [], values: [] };
         }
-        this.propValuesMap.props.keys.push(key);
-        this.propValuesMap.props.values.push({
+        this.propValuesMap['props']!['keys'].push(key);
+        this.propValuesMap['props']!['values'].push({
           start:
-            (node[
-              key as keyof Omit<CoreNodeAnimateProps, 'shaderProps'>
-            ] as number) ?? 0,
+            node[key as keyof Omit<CoreNodeAnimateProps, 'shaderProps'>] || 0,
           target: props[
             key as keyof Omit<CoreNodeAnimateProps, 'shaderProps'>
           ] as number,
           isColor: key.indexOf('color') !== -1,
         });
       } else if (key === 'shaderProps' && node.shader !== null) {
-        this.propValuesMap.shaderProps = { keys: [], values: [] };
-        for (const spKey in props.shaderProps) {
-          let start = node.shader.props![spKey];
+        this.propValuesMap['shaderProps'] = { keys: [], values: [] };
+        for (const key in props.shaderProps) {
+          let start = node.shader.props![key];
           if (Array.isArray(start) === true) {
             start = start[0];
           }
-          this.propValuesMap.shaderProps.keys.push(spKey);
-          this.propValuesMap.shaderProps.values.push({
+          this.propValuesMap['shaderProps']!['keys'].push(key);
+          this.propValuesMap['shaderProps']!['values'].push({
             start,
-            target: props.shaderProps[spKey] as number,
-            isColor: spKey.indexOf('color') !== -1,
+            target: props.shaderProps[key] as number,
+            isColor: key.indexOf('color') !== -1,
           });
         }
       }
@@ -132,16 +130,16 @@ export class CoreAnimation extends EventEmitter {
 
   restore() {
     this.reset();
-    if (this.propValuesMap.props !== null) {
+    if (this.propValuesMap['props'] !== null) {
       this.restoreValues(
         this.node as unknown as Record<string, number>,
-        this.propValuesMap.props,
+        this.propValuesMap['props'],
       );
     }
-    if (this.propValuesMap.shaderProps !== null) {
+    if (this.propValuesMap['shaderProps'] !== null) {
       this.restoreValues(
         this.node.shader!.props as Record<string, number>,
-        this.propValuesMap.shaderProps,
+        this.propValuesMap['shaderProps'],
       );
     }
   }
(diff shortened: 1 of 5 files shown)


src/core/animations/CoreAnimation.ts
```diff
@@ -55,27 +68,33 @@ export class CoreAnimation extends EventEmitter {
 
     for (const key in props) {
       if (key !== 'shaderProps') {
-        if (this.propValuesMap['props'] === undefined) {
-          this.propValuesMap['props'] = {};
```
wouterlucas, line 59: named lookups like `['props']` are faster then dot notation lookups.

src/core/animations/CoreAnimation.ts
```diff
@@ -55,27 +68,33 @@ export class CoreAnimation extends EventEmitter {
...
+        if (this.propValuesMap.props === null) {
+          this.propValuesMap.props = { keys: [], values: [] };
         }
-        this.propValuesMap['props'][key] = {
+        this.propValuesMap.props.keys.push(key);
```
wouterlucas: same here, `['props']` is faster

src/core/animations/CoreAnimation.ts
```diff
@@ -55,27 +68,33 @@ export class CoreAnimation extends EventEmitter {
...
+        this.propValuesMap.props.values.push({
           start:
-            node[key as keyof Omit<CoreNodeAnimateProps, 'shaderProps'>] || 0,
+            (node[
+              key as keyof Omit<CoreNodeAnimateProps, 'shaderProps'>
+            ] as number) ?? 0,
```
wouterlucas: why change it to `??` over an `||` ?

wouterlucas: also why `()` it?

src/core/animations/CoreAnimation.ts
```diff
@@ -55,27 +68,33 @@ export class CoreAnimation extends EventEmitter {
...
             key as keyof Omit<CoreNodeAnimateProps, 'shaderProps'>
           ] as number,
-        };
+          isColor: key.indexOf('color') !== -1,
+        });
       } else if (key === 'shaderProps' && node.shader !== null) {
-        this.propValuesMap['shaderProps'] = {};
-        for (const key in props.shaderProps) {
-          let start = node.shader.props![key];
+        this.propValuesMap.shaderProps = { keys: [], values: [] };
```
wouterlucas: no need to change the `['shaderProps']` lookup
no need to change the for loop, doesnt change anything

only the `{}` change is valid here

src/core/animations/CoreAnimation.ts
```diff
@@ -55,27 +68,33 @@ export class CoreAnimation extends EventEmitter {
...
+        for (const spKey in props.shaderProps) {
```
wouterlucas: ???

src/core/animations/CoreAnimation.ts
```diff
@@ -55,27 +68,33 @@ export class CoreAnimation extends EventEmitter {
...
+          let start = node.shader.props![spKey];
           if (Array.isArray(start) === true) {
             start = start[0];
           }
-          this.propValuesMap['shaderProps'][key] = {
```
wouterlucas, line 75: same here

src/core/animations/CoreAnimation.ts
```diff
@@ -102,56 +121,50 @@ export class CoreAnimation extends EventEmitter {
...
-
-    for (let i = 0; i < eLength; i++) {
-      const [key, value] = entries[i]!;
-      target[key] = value.start;
+  private restoreValues(target: Record<string, number>, group: PropGroup) {
+    const keys = group.keys;
+    const values = group.values;
+    const length = keys.length;
+    for (let i = 0; i < length; i++) {
+      target[keys[i]!] = values[i]!.start;
```
wouterlucas, line 127: nice!!!

src/core/animations/CoreAnimation.ts
```diff
@@ -102,56 +121,50 @@ export class CoreAnimation extends EventEmitter {
...
     }
   }
 
   restore() {
     this.reset();
-    if (this.propValuesMap['props'] !== undefined) {
```
wouterlucas, line 120: no need to change lookup method

src/core/animations/CoreAnimation.ts
```diff
@@ -102,56 +121,50 @@ export class CoreAnimation extends EventEmitter {
...
+    if (this.propValuesMap.props !== null) {
       this.restoreValues(
         this.node as unknown as Record<string, number>,
-        this.propValuesMap['props'],
+        this.propValuesMap.props,
       );
     }
-    if (this.propValuesMap['shaderProps'] !== undefined) {
+    if (this.propValuesMap.shaderProps !== null) {
```
wouterlucas: no need

src/core/animations/CoreAnimation.ts
```diff
@@ -102,56 +121,50 @@ export class CoreAnimation extends EventEmitter {
...
     }
   }
 
   restore() {
     this.reset();
-    if (this.propValuesMap['props'] !== undefined) {
+    if (this.propValuesMap.props !== null) {
       this.restoreValues(
         this.node as unknown as Record<string, number>,
-        this.propValuesMap['props'],
```
wouterlucas, line 123: no need to change it

src/core/animations/CoreAnimation.ts
```diff
@@ -102,56 +121,50 @@ export class CoreAnimation extends EventEmitter {
...
+        this.propValuesMap.props,
       );
     }
-    if (this.propValuesMap['shaderProps'] !== undefined) {
+    if (this.propValuesMap.shaderProps !== null) {
       this.restoreValues(
         this.node.shader!.props as Record<string, number>,
-        this.propValuesMap['shaderProps'],
+        this.propValuesMap.shaderProps,
```
wouterlucas: no need

src/core/animations/CoreAnimation.ts
```diff
@@ -102,56 +121,50 @@ export class CoreAnimation extends EventEmitter {
...
     }
   }
 
   reverse() {
     this.progress = 0;
 
-    if (this.propValuesMap['props'] !== undefined) {
-      this.reverseValues(this.propValuesMap['props']);
+    if (this.propValuesMap.props !== null) {
+      this.reverseValues(this.propValuesMap.props);
```
wouterlucas: no need 2x

src/core/animations/CoreAnimation.ts
```diff
@@ -102,56 +121,50 @@ export class CoreAnimation extends EventEmitter {
...
     }
-    if (this.propValuesMap['shaderProps'] !== undefined) {
-      this.reverseValues(this.propValuesMap['shaderProps']);
+    if (this.propValuesMap.shaderProps !== null) {
+      this.reverseValues(this.propValuesMap.shaderProps);
```
wouterlucas: no need 2x

src/core/animations/CoreAnimation.ts
```diff
@@ -264,17 +282,17 @@ export class CoreAnimation extends EventEmitter {
       }
     }
 
-    if (this.propValuesMap['props'] !== undefined) {
+    if (this.propValuesMap.props !== null) {
```
wouterlucas: nope

src/core/animations/CoreAnimation.ts
```diff
@@ -264,17 +282,17 @@ export class CoreAnimation extends EventEmitter {
...
       this.updateValues(
         this.node as unknown as Record<string, number>,
-        this.propValuesMap['props'],
+        this.propValuesMap.props,
```
wouterlucas: pleeeasseeeee

src/core/animations/CoreAnimation.ts
```diff
@@ -264,17 +282,17 @@ export class CoreAnimation extends EventEmitter {
...
         easing,
       );
     }
-    if (this.propValuesMap['shaderProps'] !== undefined) {
+    if (this.propValuesMap.shaderProps !== null) {
```
wouterlucas: stop it

src/core/animations/CoreAnimation.ts
```diff
@@ -264,17 +282,17 @@ export class CoreAnimation extends EventEmitter {
...
       this.updateValues(
         this.node.shader!.props as Record<string, number>,
-        this.propValuesMap['shaderProps'],
+        this.propValuesMap.shaderProps,
```
wouterlucas: bad AI

wouterlucas (changes requested):
I think your agent needs better instructions and stop messing with my value lookups 😅 

`['value']` is a lot faster on hot paths over `.value`  lookups - esp on older / constrained devices.

elsassph:
@wouterlucas 
> ['value'] is a lot faster on hot paths over .value lookups - esp on older / constrained devices.

source?

Also, I'd double check that JS minifier doesn't "optimise" for characters and replace those with dot access (I can tell that Terser and SWC do). 

PS: JS minifiers also replace `true` with `!0` which means that I'm not sure `if (a === true)` is going to be really faster (as `a === !0`) than `if (a)`. My 2 cents :)

wouterlucas:
> @wouterlucas
> 
> > ['value'] is a lot faster on hot paths over .value lookups - esp on older / constrained devices.
> 
> source?
> 
> Also, I'd double check that JS minifier doesn't "optimise" for characters and replace those with dot access (I can tell that Terser and SWC do).
> 
> PS: JS minifiers also replace `true` with `!0` which means that I'm not sure `if (a === true)` is going to be really faster (as `a === !0`) than `if (a)`. My 2 cents :)

its been a while since I ran the tests, but basically leanings from:
https://github.com/wouterlucas/jitful

different browsers apply different tactics but these worked best for our main target devices
re minifiers, good point. Is there a way to make it not do that? 😄

wouterlucas (approved):

PR #804, opened 2026-06-04
feat(vrt): add Canvas2D visual regression tests

wouterlucas:
- visual-regression: add --renderMode flag (webgl|canvas|all) to run VRT against Canvas2D renderer, storing snapshots in a separate chromium-ci-canvas/ subdirectory to keep them isolated from WebGL
- CI (tests.yml): pass --renderMode all so both renderers are always exercised on every PR and merge group check
- examples/index.ts: wrap per-test automation() calls in try/catch so an unsupported feature in one test (e.g. SDF text in canvas mode) cannot block doneTests() and hang the entire runner
- certified-snapshots/chromium-ci-canvas: 122 baseline canvas snapshots captured and committed


wouterlucas:
guess we got to check a couple of these @jfboeve 😅

PR #807, opened 2026-06-10
fix(webgl): correctly handle premultiplied-alpha for ImageBitmap

wouterlucas:
The WebGL spec permits implementations to ignore UNPACK_PREMULTIPLY_ALPHA_WEBGL for ImageBitmap sources — premultiplication must occur inside createImageBitmap(). Older WPEWebKit (Xi6 / Chrome 50 era) does not honour the 'premultiplyAlpha' option on createImageBitmap, delivering straight-alpha pixels.  Uploading those with UNPACK_PREMULTIPLY_ALPHA_WEBGL=false into a pipeline that uses blendFunc(ONE, ONE_MINUS_SRC_ALPHA) produces washed-out compositing artefacts.

Changes:
- TextureData / ImageResponse: add optional alreadyPremultiplied flag that platforms set to indicate whether the bitmap pixels are already premultiplied
- WebPlatform: sets alreadyPremultiplied: true (both crop and no-crop paths)
- WebPlatformChrome50: sets alreadyPremultiplied: false (WPEWebKit straight-alpha)
- ImageWorkerDefault: alreadyPremultiplied: true; also fix secondary bug where premultiplyAlpha resolved the raw input param instead of the computed withAlphaChannel value
- ImageWorkerNoOptions: alreadyPremultiplied: false + updated return type
- WebGlCtxTexture: replace hard-coded isImageBitmap ? false : ... with explicit branch using alreadyPremultiplied ?? true (undefined preserves old behaviour)

### DIFF
--- a/src/core/platforms/web/WebPlatform.ts
+++ b/src/core/platforms/web/WebPlatform.ts
@@ -230,7 +230,11 @@ export class WebPlatform extends Platform {
         colorSpaceConversion: 'none',
         imageOrientation: 'none',
       });
-      return { data: bitmap, premultiplyAlpha: hasAlphaChannel };
+      return {
+        data: bitmap,
+        premultiplyAlpha: hasAlphaChannel,
+        premultiplied: true,
+      };
     }
 
     // default createImageBitmap without crop but with options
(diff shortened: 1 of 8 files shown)


jfboeve (commented):
I would prefer `premultiplied` over `alreadyPremultiplied`. otherwise lgtm

wouterlucas:
> I would prefer `premultiplied` over `alreadyPremultiplied`. otherwise lgtm

done!

jfboeve (approved):

PR #817, opened 2026-06-24
Add `TextureOptions` export

guilhermesimoes:
Also update some `WebGlContextWrapper` comments


exports/index.ts
```diff
@@ -47,6 +47,7 @@ export * from '../src/common/CommonTypes.js';
 export {
   CoreTextureManager,
   type TextureMap,
+  type TextureOptions,
```
wouterlucas, line 50: do you need to export Resize Options `ResizeModeOptions` too as a type for this? for sake of completeness.

wouterlucas (approved):
small export remark, other wise LGTM

PR #818, opened 2026-06-24
Lint unused variables, params and imports

guilhermesimoes:
Noticed some unused vars, started removing them, then decided to enable the available linting rule to prevent this.

### DIFF
--- a/src/core/renderers/webgl/WebGlShaderNode.ts
+++ b/src/core/renderers/webgl/WebGlShaderNode.ts
@@ -125,6 +125,7 @@ export class WebGlShaderNode<
   }
 
   updateUniformUsage(): void {
+    /* eslint-disable @typescript-eslint/no-unused-vars */
     for (const _ in this.uniforms.single) {
       this.uniforms.hasStoredUniforms = true;
       return;
@@ -141,6 +142,7 @@ export class WebGlShaderNode<
       this.uniforms.hasStoredUniforms = true;
       return;
     }
+    /* eslint-enable @typescript-eslint/no-unused-vars */
     this.uniforms.hasStoredUniforms = false;
   }
 
--- a/src/core/renderers/webgl/WebGlShaderNode.ts
+++ b/src/core/renderers/webgl/WebGlShaderNode.ts
@@ -125,7 +125,6 @@ export class WebGlShaderNode<
   }
 
   updateUniformUsage(): void {
-    /* eslint-disable @typescript-eslint/no-unused-vars */
     for (const _ in this.uniforms.single) {
       this.uniforms.hasStoredUniforms = true;
       return;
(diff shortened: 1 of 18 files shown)


src/core/renderers/webgl/WebGlShaderNode.ts
```diff
@@ -141,6 +142,7 @@ export class WebGlShaderNode<
       this.uniforms.hasStoredUniforms = true;
       return;
     }
+    /* eslint-enable @typescript-eslint/no-unused-vars */
```
wouterlucas: probably better to update the code to be more idiomatic then adding a disable block?

src/core/renderers/webgl/internal/ShaderUtils.ts
```diff
@@ -79,6 +79,7 @@ export interface SupportedSetUniforms {
     | 'uniform4i';
 }
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 type SupportSetUniforms =
```
wouterlucas: I think this is dead code? we can just remove it instead of disabling the linting?
@jfboeve ?

wouterlucas (changes requested):

src/core/renderers/webgl/internal/ShaderUtils.ts
guilhermesimoes: Yeah, that's fine by me if that's the route you want to take

src/core/renderers/webgl/WebGlShaderNode.ts
guilhermesimoes: This was added recently in #798

I would have written it as:

```ts
this.uniforms.hasStoredUniforms =
  Object.keys(this.uniforms.single).length > 0 ||
  Object.keys(this.uniforms.vec2).length > 0 ||
  Object.keys(this.uniforms.vec3).length > 0 ||
  Object.keys(this.uniforms.vec4).length > 0;
```

But I would imagine this is probably slower, not sure.

What do you think?

src/core/renderers/webgl/internal/ShaderUtils.ts
wouterlucas: if its really dead - yeah just yoink it out 😄

src/core/renderers/webgl/WebGlShaderNode.ts
jfboeve: This is what I had before, however this creates a new array for each Object.keys, which is why I went for the seperate loops

src/core/renderers/webgl/internal/ShaderUtils.ts
guilhermesimoes: Done!

jfboeve (dismissed):

src/core/renderers/webgl/WebGlShaderNode.ts
guilhermesimoes: Yeah I thought it'd be something like that.

In the meantime I found that I could configure the rule with `varsIgnorePattern: '^_'`. So I've added that option and removed the eslint-disable from this file.

Does this look ok now?

jfboeve (approved):

wouterlucas (approved):

PR #830, opened 2026-07-01
feat: added premultiplyAlpha parameter to uniformRGBA

jfboeve:
Added alpha premultiplication parameter to uniformRGBA in the WebGlShaderNode / WebGlShaderType.

This parameter is a boolean and the default is false.

### DIFF
--- a/src/core/renderers/webgl/WebGlShaderNode.ts
+++ b/src/core/renderers/webgl/WebGlShaderNode.ts
@@ -161,7 +161,10 @@ export class WebGlShaderNode<
     }
 
     const a = rgba[3];
-    this.uniform4fa(location, [rgba[0] * a, rgba[1] * a, rgba[2] * a, a]);
+    rgba[0] *= a;
+    rgba[1] *= a;
+    rgba[2] *= a;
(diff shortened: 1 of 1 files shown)


wouterlucas (approved):

src/core/renderers/webgl/WebGlShaderNode.ts
```diff
@@ -148,12 +148,24 @@ export class WebGlShaderNode<
...
+    if (premultiplyAlpha === false || rgba[3] === 1) {
+      this.uniform4fa(location, rgba);
+      return;
+    }
+
+    const a = rgba[3];
+    rgba[0] *= a;
+    rgba[1] *= a;
+    rgba[2] *= a;
+    this.uniform4fa(location, rgba);
```
guilhermesimoes, line 168: What about:
```suggestion
    if (premultiplyAlpha) {
      const a = rgba[3];
      if (a !== 1) {
        rgba[0] *= a;
        rgba[1] *= a;
        rgba[2] *= a;
      }
    }
    this.uniform4fa(location, rgba);
```
?

PR #831, opened 2026-07-07
Improve canvas text

eXecutable:
`getImageData` forces the browser to copy the pixels drawn by the canvas to JS world. By using ImageBitmap instead the pixels can be drawn on the GPU and stay there.


wouterlucas:
Nice thanks!! we should also add support for this in the Platform layers right? else we would never utilize offscreen canvas properly unless you have your own custom platform.

eXecutable:
> Nice thanks!! we should also add support for this in the Platform layers right? else we would never utilize offscreen canvas properly unless you have your own custom platform.

Updated WebPlatform, not sure that is what you meant.

PR #843, opened 2026-07-24
Handle possibly undefined letter spacing

guilhermesimoes:
We have some code that is doing:
```ts
textElement.letterSpacing = config.letterSpacing;
```
where `config.letterSpacing` can be `undefined`.

Now, I know, we need to fix this on our side and do something like:
```ts
textElement.letterSpacing = config.letterSpacing || 0;
```
And we've already done this.

However, this works in L2, so I thought we might as well make the L3 text layout engine a little more robust and handle it as well.

### DIFF
--- a/src/core/text-rendering/CanvasFontHandler.ts
+++ b/src/core/text-rendering/CanvasFontHandler.ts
@@ -225,7 +225,7 @@ export const processFontMetrics = (
 
 export const measureText = (
   text: string,
-  fontFamily: string,
+  _fontFamily: string, // Unused, but necessary to match SdfFontHandler's signature
   letterSpacing: number,
 ) => {
   if (letterSpacing === 0) {
--- a/src/core/text-rendering/CanvasTextRenderer.ts
+++ b/src/core/text-rendering/CanvasTextRenderer.ts
@@ -147,7 +147,7 @@ const renderText = (props: CoreTextNodeProps): TextRenderInfo => {
 
   const metrics = CanvasFontHandler.getFontMetrics(fontFamily, fontSize);
 
-  const letterSpacing = props.letterSpacing;
+  const letterSpacing = props.letterSpacing || 0;
 
   const [
     lines,
--- a/src/core/text-rendering/SdfTextRenderer.ts
+++ b/src/core/text-rendering/SdfTextRenderer.ts
@@ -252,7 +252,7 @@ const generateTextLayout = (
     lines,
     remainingLines,
     hasRemainingText,
-    bareLineHeight,
+    _bareLineHeight,
     lineHeightPx,
     effectiveWidth,
     effectiveHeight,
@@ -270,9 +270,6 @@ const generateTextLayout = (
     maxWidth,
     maxHeight,
   );
-
-  // Suppress unused-variable warning for bareLineHeight (consumed by mapTextLayout tuple).
-  void bareLineHeight;
 
   // --- Pre-compute decoration offsets in design-unit space ---
   // commonFontData.base is the BMFont "base" value: the y-distance from the top of the


jfboeve:
can't you just do this on line 150:

```js
  const letterSpacing = props.letterSpacing || 0;
```

guilhermesimoes:
Yeah, we can. Would that be preferable?

guilhermesimoes:
Btw Opus 4.6 had the following to say:

> `!variable` will be marginally faster, especially on older engines.
>
> - `=== 0` requires a strict equality check: verify the type is number, then compare the value to 0. Two steps.
> - `!variable` is a single falsy check — at the machine level, it's essentially just testing if the value is zero.
>
> On modern JIT engines (V8, SpiderMonkey), both compile to the same machine instruction once the type is known, so the difference vanishes. On older/non-JIT engines, `!variable` skips the explicit type check, giving it a slight edge.

I asked "Isn't `!variable` a form of coercion, and isn't that slower than verifying if the type is number?" and it replied:

> You raise a fair point. Yes, `!variable` involves type coercion (ToBoolean), which in theory adds overhead. However, for numbers specifically, the ToBoolean spec is trivial — it's just "is it 0 or NaN?" — which maps directly to a zero-check at the machine level.
> 
> Meanwhile, `=== 0` with strict equality also has work to do: it checks that both operands share the same type, then compares values.
> 
> In practice, both paths end up at roughly the same cost: a single comparison instruction. The coercion in `!variable` isn't the expensive kind (like string-to-number); it's a simple branch on zero.
> 
> Bottom line: The difference is so small it's unlikely to be measurable even in a hot loop. If you're optimizing at this level for `letterSpacing`, you'd get more out of profiling the actual rendering path. Use whichever reads better to you — for a known-number, both are effectively equivalent.

😅

jfboeve:
Would prefer adhering to the guidelines wouter has setup:
https://github.com/lightning-js/renderer/blob/main/.github/copilot-instructions.md

guilhermesimoes:
Ah! Ok, I had never read that. Let me update the PR then

guilhermesimoes:
Updated!

jfboeve (approved):

wouterlucas:
> Btw Opus 4.6 had the following to say:
> 
> > `!variable` will be marginally faster, especially on older engines.
> > 
> > * `=== 0` requires a strict equality check: verify the type is number, then compare the value to 0. Two steps.
> > * `!variable` is a single falsy check — at the machine level, it's essentially just testing if the value is zero.
> > 
> > On modern JIT engines (V8, SpiderMonkey), both compile to the same machine instruction once the type is known, so the difference vanishes. On older/non-JIT engines, `!variable` skips the explicit type check, giving it a slight edge.
> 
> I asked "Isn't `!variable` a form of coercion, and isn't that slower than verifying if the type is number?" and it replied:
> 
> > You raise a fair point. Yes, `!variable` involves type coercion (ToBoolean), which in theory adds overhead. However, for numbers specifically, the ToBoolean spec is trivial — it's just "is it 0 or NaN?" — which maps directly to a zero-check at the machine level.
> > Meanwhile, `=== 0` with strict equality also has work to do: it checks that both operands share the same type, then compares values.
> > In practice, both paths end up at roughly the same cost: a single comparison instruction. The coercion in `!variable` isn't the expensive kind (like string-to-number); it's a simple branch on zero.
> > Bottom line: The difference is so small it's unlikely to be measurable even in a hot loop. If you're optimizing at this level for `letterSpacing`, you'd get more out of profiling the actual rendering path. Use whichever reads better to you — for a known-number, both are effectively equivalent.
> 
> 😅

that said your AI is wrong 😅 

a fallsy check, checks for a series of conditions, like is it `false`, is it `undefined`, is it `null`, etc - so essentially it executes more then 1 comparison. Where strict checks are a single comparison, type casting is faster then multiple `IFs`. In modern engines this is negligible, yet on older engines (and those we target) this matters. It's only a little bit, but all combined they add up over a large scope of a project esp on hot paths.

Same goes with dot notations and multiple object property access, nesting, this scope creation on functions and forEach/map/filter for loops. All though all marginally they do matter at scale on limited resources on older browsers with half-ass JIT implementations 😄 

Basically the old school basic JavaScript will always out perform the modern fancy JavaScript on these devices.

PR #846, opened 2026-08-04
fix: possible memory leak in inspector code

AndresMonk:


src/main-api/Inspector.ts
```diff
@@ -962,8 +962,41 @@ export class Inspector {
         });
         coreNodeListeners.clear();
 
-        this.destroyNode(node.id);
+        // Clean up activeAnimations entries for this node
+        Inspector.activeAnimations.forEach((anim, id) => {
```
michielvandergeest: maybe we should change this to using a for loop instead of foreach?

src/main-api/Inspector.ts
```diff
@@ -962,8 +962,41 @@ export class Inspector {
...
+        // Break circular references between node and div
+        (div as any).node = null;
+        (node as any).div = null;
+
+        // Call original destroy (recursively destroys renderer children)
         originalDestroy.call(node);
+
+        // Remove all defineProperty traps to release closures that capture
+        // node, div, originalProp, and the Inspector instance
+        knownProperties.forEach((property) => {
```
michielvandergeest: same here, for instead of foreach

wouterlucas (approved):
* chef's kiss *

PR #848, opened 2026-08-05
perf: reduce GC pressure and hot-path overhead

wouterlucas:
## Summary

A set of focused performance optimisations targeting GC pressure and CPU overhead in hot paths on embedded devices.

### Changes

**EventEmitter (lazy allocation + hidden-class preservation)**
- `eventListeners` map starts as `null`, allocated on first `on()` call — saves one `{}` per instance for nodes/textures that never have listeners
- `off()` uses `= undefined` instead of `delete` to avoid V8 hidden-class transitions
- Added `hasListeners(event?)` helper for efficient presence checks

**CoreNode.setUpdateType (parent traversal short-circuit)**
- When parent already has the `Children` flag set, skip the recursive call up the parent chain — prevents redundant traversal when many siblings dirty the same parent in a single frame

**CoreNode.clippingRect (lazy allocation)**
- Non-clipping nodes (the majority) now share a frozen `NO_CLIPPING_RECT` singleton instead of each allocating a mutable rect object at construction

**Texture.renderableOwners (Array → Set)**
- O(1) `has/add/delete` replaces O(n) `indexOf/splice/push` — significant improvement for textures with many owners or frequent ownership changes

**CoreTextureManager (inverseKeyCache removal)**
- Replaced `WeakMap<Texture, string>` with a direct `cacheKey` property on Texture — eliminates WeakMap lookup overhead and one allocation per cached texture

### Testing

- All 417 unit tests pass
- Clean TypeScript build (no type errors)
- No public API surface changes

### DIFF
--- a/src/common/EventEmitter.ts
+++ b/src/common/EventEmitter.ts
@@ -23,26 +23,44 @@ import type { IEventEmitter } from './IEventEmitter.js';
 export type EventListener = (target: any, data: any) => void;
 /**
  * EventEmitter base class
+ *
+ * @remarks
+ * The `eventListeners` map is allocated lazily on the first `on()` call.
+ * Many objects (e.g. CoreNodes that never have listeners attached) avoid
+ * the cost of allocating an empty `{}` at construction time.
  */
 export class EventEmitter implements IEventEmitter {
-  protected eventListeners: { [eventName: string]: EventListener[] } = {};
+  private eventListeners: {
+    [eventName: string]: EventListener[] | undefined;
+  } | null = null;
 
   on(event: string, listener: EventListener): void {
-    let listeners = this.eventListeners[event];
-    if (!listeners) {
+    let map = this.eventListeners;
+    if (map === null) {
+      map = {};
+      this.eventListeners = map;
+    }
+    let listeners = map[event];
+    if (listeners === undefined) {
       listeners = [];
-      this.eventListeners[event] = listeners;
+      map[event] = listeners;
     }
     listeners.push(listener);
   }
 
   off(event: string, listener?: EventListener): void {
-    const listeners = this.eventListeners[event];
-    if (!listeners) {
+    const map = this.eventListeners;
+    if (map === null) {
       return;
     }
-    if (!listener) {
-      delete this.eventListeners[event];
+    const listeners = map[event];
+    if (listeners === undefined) {
+      return;
+    }
+    if (listener === undefined) {
+      // Set to undefined instead of `delete` to avoid V8 hidden-class
+      // transitions on the eventListeners object.
+      map[event] = undefined;
       return;
     }
     const index = listeners.indexOf(listener);
@@ -60,7 +78,11 @@ export class EventEmitter implements IEventEmitter {
   }
 
   emit(event: string, data?: any): void {
-    const listeners = this.eventListeners[event];
+    const map = this.eventListeners;
+    if (map === null) {
+      return;
+    }
+    const listeners = map[event];
     if (listeners === undefined || listeners.length === 0) {
       return;
     }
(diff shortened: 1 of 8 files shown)


jfboeve (approved):

src/common/EventEmitter.ts
```diff
@@ -23,26 +23,44 @@ import type { IEventEmitter } from './IEventEmitter.js';
...
+ * @remarks
+ * The `eventListeners` map is allocated lazily on the first `on()` call.
+ * Many objects (e.g. CoreNodes that never have listeners attached) avoid
+ * the cost of allocating an empty `{}` at construction time.
  */
 export class EventEmitter implements IEventEmitter {
-  protected eventListeners: { [eventName: string]: EventListener[] } = {};
+  private eventListeners: {
+    [eventName: string]: EventListener[] | undefined;
+  } | null = null;
```
guilhermesimoes, line 35:
```suggestion
  private declare eventListeners?: {
    [eventName: string]: EventListener[] | undefined;
  };
```
Would this be an improvement? It would prevent `eventListeners` from even being declared.

(The only drawback I see is that nodes where `eventListeners` are set will have a different object shape. But imagine those would be in a minority)

src/common/EventEmitter.ts
```diff
@@ -23,26 +23,44 @@ import type { IEventEmitter } from './IEventEmitter.js';
...
+      map = {};
+      this.eventListeners = map;
+    }
+    let listeners = map[event];
+    if (listeners === undefined) {
       listeners = [];
-      this.eventListeners[event] = listeners;
+      map[event] = listeners;
     }
     listeners.push(listener);
```
guilhermesimoes, line 48: Btw this could be rewritten like so:
```ts
let map = this.eventListeners ||= {};
let listeners = map[event] ||= [];
listeners.push(listener);
```
Or even in a single line:
```ts
((this.eventListeners ||= {})[event] ||= []).push(listener)
```
😅 
But not sure you can use `||=`, it's only available since `ES2021`. We transpile Lightning itself, but not sure other clients to the same.

src/common/EventEmitter.ts
wouterlucas: yea the idea is we dont create the set or map until there are actually listeners. 

It reduces spawning constant structures without anyone actually using them and shifts the creation impact to the first subscriber

src/common/EventEmitter.ts
wouterlucas: what benefit would that give?

src/common/EventEmitter.ts
guilhermesimoes: Yeah, I get that, makes total sense!

Do note that the current code gets transpiled to this:
```ts
export class EventEmitter {
    constructor() {
        this.eventListeners = null;
    }
}
```
[Playground](https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBAYwDYEMDOa4FEBuwB2MWAtgJYwzBRwDeAUHHGFKTipXMHoQDKlqV8VNAC5aDRnADaXAjAByKYsDECW+AOYBdMbjl8BBKlK1wAPnACu+ACbAAZqSE2A3BIC+5uPktIkcAF5vXyQ3dyA)

My suggestion was to use `declare` + optional property so the code gets transpiled to this:

```ts
export class EventEmitter {}
```

[Playground](https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBAYwDYEMDOa4FEBuwB2MWAtgJYwzBRwDeAUHHGFKTipXACbDIpTBxgeQgBlSaSvipoA-AC5aDRnADaQgjAByKYsAUSW+AOYBdBbg1iJBKipNwAPnACu+bgDNSUzgG4lAXz9-IA)

What do you think?

src/common/EventEmitter.ts
guilhermesimoes: Just more terse, succinct. `||=` is a common idiom in rubyland, so maybe that's why I like it 😄 
But it's still uncommon in jsland

philippe-wm:
Beyond memory pressure, I believe V8 prefers when the shape of objects don't change (which adding a field do)

PR #856, opened 2026-08-06
fix: memory leak in core node

AndresMonk:


wouterlucas (dismissed):
nice find - thanks Andres!

wouterlucas (approved):
thanks!

PR #859, opened 2026-08-06
refactor: creating reusable mock stage object

AndresMonk:
Extracting out the stage mocking process to make it reusable rather than declared in each file


wouterlucas:
don't have too much fun now @AndresMonk 😄

wouterlucas (approved):

PR #864, opened 2026-08-07
Iterate over children with for loop inside `destroy`

guilhermesimoes:
This is a super rare edge case but due to a programming mistake a `destroyed` node may be re-attached.
Calling `destroy` on that node's parent would freeze the app due to the combination of the while loop and `destroy`'s top guard.
https://github.com/lightning-js/renderer/blob/345ca62ca00e1b4241603ce4def41872c24d15e7/src/core/CoreNode.ts#L2020-L2023
https://github.com/lightning-js/renderer/blob/345ca62ca00e1b4241603ce4def41872c24d15e7/src/core/CoreNode.ts#L2003-L2006

### DIFF
--- a/src/core/CoreNode.ts
+++ b/src/core/CoreNode.ts
@@ -2017,12 +2017,12 @@ export class CoreNode extends EventEmitter {
       this.stage.untrackTimedNode(this);
     }
 
+    const { children, parent } = this;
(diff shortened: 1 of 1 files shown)


wouterlucas (approved):
LGTM - thanks!

PR #865, opened 2026-08-07
perf: sort main-pass SDF text writes for fewer draw calls

wouterlucas:
## What

Defer main-pass SDF text rendering into a scene-wide sorted list instead of writing interleaved glyph runs inline during the traversal.

SDF text already batches into a shared GPU buffer, but only *consecutive* writes sharing a vertex layout (plain 6f vs rich 7f), glyph atlas (per font family) and scissor clip rect merge into one `drawElements`. Apps that interleave plain/rich text (cards, rails) forced a draw call per region. This sorts the visible SDF text nodes by batch key at render-list build time so each (layout, atlas, clip) group becomes one contiguous range and merges into a single draw call, with zero app-side grouping.

## Changes

- `lib/utils.ts` — new `SdfBatchKey` type and `compareSdfBatchKeys` total ordering (matches the merge key in `WebGlRenderer.finalizeSdfBatch`: layout, atlas, clip rect incl. radius).
- `CoreNode.getSdfBatchKey()` — virtual returning `null`; `CoreTextNode` overrides it to return the key for main-pass SDF text, `null` for canvas text, un-laid-out text, and text inside render-to-texture subtrees.
- `Stage` — `buildRenderList` collects main-pass SDF text nodes (scene order) instead of writing inline; `sortRenderListSdfTextNodes` sorts them with a stable sort (equal keys keep scene order / relative z), cached across frames; `writeRenderListSdfText` emits them after the main list on every frame. RTT traversal and all non-SDF nodes keep the inline path.
- Tests — 6 `getSdfBatchKey` tests in `CoreTextNode.test.ts`; 12 Stage tests covering comparator order, deferral, sorted emission, inline paths (images, rounded clips, RTT), visibility skips, stable equal-keys, and cached replay.

## Correctness

The sort is best-effort: `finalizeSdfBatch` still keys off actual write-time values, so a stale cached order (e.g. a `fontFamily`/clip change without a structural rebuild) only costs extra draw calls, never wrong output.

## Verification

- `pnpm test` 272/272, `tsc --build`, prettier, eslint clean.
- SDF visual regression suite 21/21 PASS (`chromium-local`).
- Stress A/B example: page 1 (interleaved) vs page 2 (grouped) differ by 20/921,600 px (0.002%) — sorted rendering merges interleaved text into the same grouped output.


wouterlucas:
Rebased onto latest `main` (`3aef7251`), which includes the merged #860 (static nodeCoords VBO / reduced vertex stride) and #861 (surgical dirty-quad uploads). Conflicts in `WebGlRenderer.ts` and `Stage.ts` resolved by keeping both: upstream quad uploads + the batched SDF pipeline. Added a stub for the new SDF buffer fields in the upstream dirty-quad test harness (the `Object.create`-based mock skips class-field initializers). All 296 unit tests pass; `tsc --build` clean.

jfboeve (approved):

PR #869, opened 2026-08-31
Make `CanvasRenderer` treat clearColor as RGBA, like `WebGlRenderer`

guilhermesimoes:
Currently if we do `stage.setClearColor(0x000000FF)` in WebGL the background is black, but in Canvas it's white (because it thinks we're using ARGB instead of RGBA). Let's change this so that both WebGL and Canvas are consistent. And also so that `app.props.color = 0x000000FF` is consistent with `app.stage.setClearColor(0x000000FF)`.

### DIFF
--- a/src/core/lib/colorCache.ts
+++ b/src/core/lib/colorCache.ts
@@ -3,18 +3,24 @@ import { parseToAbgrString, parseToRgbaString } from './colorParser.js';
 const parsedArgbColors: Map<number, string> = new Map();
 const parsedRgbaColors: Map<number, string> = new Map();
 
-export function normalizeCanvasColor(color: number, isRGBA: boolean = true) {
-  let targetCache = isRGBA === true ? parsedRgbaColors : parsedArgbColors;
-  let out = targetCache.get(color);
+export function normalizeCanvasColor(rgbaColor: number) {
+  let out = parsedRgbaColors.get(rgbaColor);
   if (out !== undefined) {
     return out;
   }
 
-  if (isRGBA === true) {
-    out = parseToRgbaString(color);
-  } else {
-    out = parseToAbgrString(color);
+  out = parseToRgbaString(rgbaColor);
+  parsedRgbaColors.set(rgbaColor, out);
+  return out;
+}
+
+export function normalizeCanvasColorArgb(argbColor: number) {
+  let out = parsedArgbColors.get(argbColor);
+  if (out !== undefined) {
+    return out;
   }
-  targetCache.set(color, out);
+
+  out = parseToAbgrString(argbColor);
+  parsedArgbColors.set(argbColor, out);
   return out;
 }
--- a/src/core/lib/colorCache.ts
+++ b/src/core/lib/colorCache.ts
@@ -3,7 +3,7 @@ import { parseToAbgrString, parseToRgbaString } from './colorParser.js';
 const parsedArgbColors: Map<number, string> = new Map();
 const parsedRgbaColors: Map<number, string> = new Map();
 
-export function normalizeCanvasColor(color: number, isRGBA: boolean = false) {
+export function normalizeCanvasColor(color: number, isRGBA: boolean = true) {
   let targetCache = isRGBA === true ? parsedRgbaColors : parsedArgbColors;
   let out = targetCache.get(color);
   if (out !== undefined) {
--- a/src/core/lib/colorParser.ts
+++ b/src/core/lib/colorParser.ts
@@ -80,6 +80,6 @@ export function parseColorRgba(rgba: number): IParsedColor {
 /**
  * Format a parsed color into a rgba CSS color
  */
-export function formatRgba({ a, r, g, b }: IParsedColor): string {
-  return `rgba(${r},${g},${b},${a})`;
+export function formatRgba(color: IParsedColor): string {
+  return `rgba(${color.r},${color.g},${color.b},${color.a})`;
 }
--- a/src/core/lib/colorParser.ts
+++ b/src/core/lib/colorParser.ts
@@ -41,25 +41,25 @@ export function parseColor(abgr: number): IParsedColor {
     return WHITE;
   }
   const a = ((abgr >>> 24) & 0xff) / 255;
-  const b = (abgr >>> 16) & 0xff & 0xff;
-  const g = (abgr >>> 8) & 0xff & 0xff;
-  const r = abgr & 0xff & 0xff;
+  const b = (abgr >>> 16) & 0xff;
+  const g = (abgr >>> 8) & 0xff;
+  const r = abgr & 0xff;
   return { isWhite: false, a, r, g, b };
 }
 
 export function parseToAbgrString(abgr: number): string {
   const a = ((abgr >>> 24) & 0xff) / 255;
-  const b = (abgr >>> 16) & 0xff & 0xff;
-  const g = (abgr >>> 8) & 0xff & 0xff;
-  const r = abgr & 0xff & 0xff;
+  const b = (abgr >>> 16) & 0xff;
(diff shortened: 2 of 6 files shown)


src/core/lib/colorCache.ts
```diff
@@ -3,7 +3,7 @@ import { parseToAbgrString, parseToRgbaString } from './colorParser.js';
 const parsedArgbColors: Map<number, string> = new Map();
 const parsedRgbaColors: Map<number, string> = new Map();
 
-export function normalizeCanvasColor(color: number, isRGBA: boolean = false) {
+export function normalizeCanvasColor(color: number, isRGBA: boolean = true) {
```
guilhermesimoes: Not sure if you agree with the second commit that inverts this default.
If you don't agree I can undo this change.

guilhermesimoes: Btw, I also thought about splitting this function into two: `normalizeCanvasColor` and `normalizeCanvasColorArgb`. I generally subscribe to the idea of [avoiding boolean params in functions](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap). Tell me what you think.

wouterlucas: I'm much more in favor of splitting it in two sep functions! Since it already gets the value from 2 distinct caches and two distinct `parseTo<variant>String` functions, it would be cheaper to make 2 explicit functions for this.

I would also immediately inline the `parseToRgbaString` / `parsedArgbColors` functions to avoid the additional function hop. And remove them from utils, no one else but this function seems to be using them.

Great suggestion and love the article!

wouterlucas (changes requested):

src/core/lib/colorCache.ts
guilhermesimoes: I split the functions into two.

The utils `parseTo<variant>String` would actually be useful to us if they were exported. I noticed we have our own implementations of these.

src/core/lib/colorParser.ts
```diff
@@ -41,25 +41,25 @@ export function parseColor(abgr: number): IParsedColor {
     return WHITE;
   }
   const a = ((abgr >>> 24) & 0xff) / 255;
-  const b = (abgr >>> 16) & 0xff & 0xff;
-  const g = (abgr >>> 8) & 0xff & 0xff;
-  const r = abgr & 0xff & 0xff;
```
guilhermesimoes, line 46: My AI agent flagged this issue: applying `& 0xff` twice is a no-op.

src/core/lib/colorParser.ts
```diff
@@ -41,25 +41,25 @@ export function parseColor(abgr: number): IParsedColor {
...
+  const g = (abgr >>> 8) & 0xff;
+  const r = abgr & 0xff;
   return { isWhite: false, a, r, g, b };
 }
 
 export function parseToAbgrString(abgr: number): string {
   const a = ((abgr >>> 24) & 0xff) / 255;
-  const b = (abgr >>> 16) & 0xff & 0xff;
-  const g = (abgr >>> 8) & 0xff & 0xff;
-  const r = abgr & 0xff & 0xff;
```
guilhermesimoes, line 54: Is this code correct?
The variable is named `abgr` and the variables are `a`, `b`, `g`, `r`.
Shouldn't it be `argb` and the variables be `a`, `r`, `g`, `b`?

wouterlucas: the variable is probably off - but the logic below works, its just the declaration of the incoming number.

src/core/lib/colorParser.ts
wouterlucas: yea that seems right

src/core/lib/colorCache.ts
```diff
@@ -3,18 +3,24 @@ import { parseToAbgrString, parseToRgbaString } from './colorParser.js';
...
+}
+
+export function normalizeCanvasColorArgb(argbColor: number) {
+  let out = parsedArgbColors.get(argbColor);
+  if (out !== undefined) {
+    return out;
   }
-  targetCache.set(color, out);
+
+  out = parseToAbgrString(argbColor);
```
wouterlucas, line 23: small nit picky thing - instead of calling `parseToAbgrString`  here

can you just inline/copy the entire contents of the function here? no need to do the function call hop and object creation when this is the only place that uses the function.

src/core/lib/colorCache.ts
```diff
@@ -3,18 +3,24 @@ import { parseToAbgrString, parseToRgbaString } from './colorParser.js';
...
+  let out = parsedRgbaColors.get(rgbaColor);
   if (out !== undefined) {
     return out;
   }
 
-  if (isRGBA === true) {
-    out = parseToRgbaString(color);
-  } else {
-    out = parseToAbgrString(color);
+  out = parseToRgbaString(rgbaColor);
```
wouterlucas, line 12: small nit picky thing - instead of calling parseToRgbaString here

can you just inline/copy the entire contents of the function here? no need to do the function call hop and object creation when this is the only place that uses the function.

src/core/lib/colorParser.ts
```diff
@@ -41,25 +41,25 @@ export function parseColor(abgr: number): IParsedColor {
...
+  const b = (abgr >>> 16) & 0xff;
+  const g = (abgr >>> 8) & 0xff;
+  const r = abgr & 0xff;
   return { isWhite: false, a, r, g, b };
 }
 
 export function parseToAbgrString(abgr: number): string {
```
wouterlucas, line 50: we should remove this from colorParser.ts and inline in the above function.

src/core/lib/colorParser.ts
```diff
@@ -41,25 +41,25 @@ export function parseColor(abgr: number): IParsedColor {
...
-  const b = (abgr >>> 16) & 0xff & 0xff;
-  const g = (abgr >>> 8) & 0xff & 0xff;
-  const r = abgr & 0xff & 0xff;
+  const b = (abgr >>> 16) & 0xff;
+  const g = (abgr >>> 8) & 0xff;
+  const r = abgr & 0xff;
   return `rgba(${r},${g},${b},${a})`;
 }
 
 export function parseToRgbaString(rgba: number): string {
```
wouterlucas, line 58: same here, just remove this entire function and inline it above.

wouterlucas (changes requested):

src/core/lib/colorParser.ts
guilhermesimoes: You sure it's correct? Shouldn't the `r` and the `b` be swapped? Like so
```ts
const a = ((argb >>> 24) & 0xff) / 255;
const r = (argb >>> 16) & 0xff;
const g = (argb >>> 8) & 0xff;
const b = argb & 0xff;
```

wouterlucas: yeah for making it look better, for sure!

guilhermesimoes: Notice I did not just swap the lines. The current code is like this:
```ts
const a = ((abgr >>> 24) & 0xff) / 255;
const b = (abgr >>> 16) & 0xff;
const g = (abgr >>> 8) & 0xff;
const r = abgr & 0xff;
```
Swapping the lines would be like so:
```ts
const a = ((abgr >>> 24) & 0xff) / 255;
const r = abgr & 0xff;
const g = (abgr >>> 8) & 0xff;
const b = (abgr >>> 16) & 0xff;
```
But isn't this code incorrect? Shouldn't it be like so:
```ts
const a = ((argb >>> 24) & 0xff) / 255;
const r = (argb >>> 16) & 0xff;
const g = (argb >>> 8) & 0xff;
const b = argb & 0xff;
```
?

PR #893, opened 2026-09-10
Fix SDF `loadFont` rejecting promise with target instead of error

guilhermesimoes:


src/core/text-rendering/SdfFontHandler.ts
```diff
@@ -355,7 +355,7 @@ export const loadFont = async (
         resolve();
       });
 
-      atlasTexture.on('failed', (error: Error) => {
+      atlasTexture.on('failed', (_target, error: Error) => {
```
guilhermesimoes, line 358: Btw, this has been an issue in our migration from L2 to L3.

Given that in none of the listeners we ever use the first argument (`target`), would it be possible to change this:
https://github.com/lightning-js/renderer/blob/3015383f114bb9b04dd186cca9bb2d973c92ce5b/src/common/EventEmitter.ts#L93
to simply

```ts
listeners[i]!(data); 
```

?
In our app we have never needed access to the target.

Usually we do:
```ts
this.on(someEvent => {
  // here we have access to `this`
});
```

Or we do:

```ts
const child = this.child;
child.on(someEvent => {
  // here we have access to `child`
});
```

Don't see why we have to receive the target as the first argument.

jfboeve: maybe we can implement this for v4.0 @wouterlucas? internally in the renderer I don't encounter use of target either?

jfboeve (approved):

src/core/text-rendering/SdfFontHandler.ts
wouterlucas: @jfboeve sure!

